PR #1, opened 2016-03-29
DAO changes

shyamakirala:


db/create_tables.sql
```diff
@@ -1,4 +1,4 @@
-CREATE TABLE IF NOT EXISTS `audit` (
+CREATE TABLE IF NOT EXISTS `audit_records` (
```
yogeshnachnani: did this file exist in master before this? I thought this would show up as a new file.

db/create_tables.sql
```diff
@@ -55,6 +55,8 @@ CREATE TABLE IF NOT EXISTS `state_machines` (
   `id` BIGINT(20) NOT NULL AUTO_INCREMENT,
   `name` VARCHAR(255) NOT NULL,
   `version` TINYINT UNSIGNED NOT NULL,
+  `start_state_id` BIGINT(20) NOT NULL,
```
yogeshnachnani: we don't need this

db/create_tables.sql
```diff
@@ -55,6 +55,8 @@ CREATE TABLE IF NOT EXISTS `state_machines` (
   `id` BIGINT(20) NOT NULL AUTO_INCREMENT,
   `name` VARCHAR(255) NOT NULL,
   `version` TINYINT UNSIGNED NOT NULL,
```
yogeshnachnani: TINYINT?

db/create_tables.sql
```diff
@@ -55,6 +55,8 @@ CREATE TABLE IF NOT EXISTS `state_machines` (
...
+  `start_state_id` BIGINT(20) NOT NULL,
+  `description` VARCHAR(300) DEFAULT NULL,
   `created_at` DATETIME(3) DEFAULT NULL,
```
yogeshnachnani: This should be DEFAULT NOW()

db/create_tables.sql
```diff
@@ -55,6 +55,8 @@ CREATE TABLE IF NOT EXISTS `state_machines` (
...
   `updated_at` DATETIME(3) DEFAULT NULL,
```
yogeshnachnani: This should be ON UPDATE CURRENT_TIMESTAMP()

yogeshnachnani: Why are we using DATETIME(3) ?

db/create_tables.sql
```diff
@@ -69,10 +71,12 @@ CREATE TABLE IF NOT EXISTS `states` (
   `id` BIGINT(20) NOT NULL AUTO_INCREMENT,
   `name` VARCHAR(255) NOT NULL,
   `version` TINYINT UNSIGNED NOT NULL,
+  `description` VARCHAR(300) DEFAULT NULL,
   `state_machine_id` BIGINT(20) NOT NULL,
   `dependencies` VARCHAR(1000) DEFAULT NULL,
```
yogeshnachnani: What are dependencies?

db/create_tables.sql
```diff
@@ -69,10 +71,12 @@ CREATE TABLE IF NOT EXISTS `states` (
...
-  `hook_class_name` VARCHAR(255) DEFAULT NULL,
-  `task_class_name` VARCHAR(255) NOT NULL,
+  `on_exit_hook` varchar(500) DEFAULT NULL,
+  `task_class` VARCHAR(500) NOT NULL,
+  `on_exit_hook` varchar(500) DEFAULT NULL,
   `retry_count` TINYINT UNSIGNED DEFAULT NULL,
```
yogeshnachnani: why tinyint?

db/create_tables.sql
```diff
@@ -39,7 +39,7 @@ CREATE TABLE IF NOT EXISTS `events` (
   `type` VARCHAR(100) NOT NULL,
   `status` VARCHAR(100) DEFAULT NULL,
   `state_machine_instance_id` VARCHAR(255) NOT NULL,
```
yogeshnachnani: why have a VARCHAR() without having any referential integrity?
Why not link it with PK of `state_machine` table

db/create_tables.sql
```diff
@@ -39,7 +39,7 @@ CREATE TABLE IF NOT EXISTS `events` (
...
-  `data` LONGTEXT,
+  `event_data` BLOB,
   `event_source` VARCHAR(100) DEFAULT NULL,
   `created_at` DATETIME(3) DEFAULT NULL,
   `updated_at` DATETIME(3) DEFAULT NULL,
```
yogeshnachnani: we don't need updated at for events

db/create_tables.sql
yogeshnachnani: do we need an auto increment ID in this table?

db/create_tables.sql
```diff
@@ -1,4 +1,4 @@
-CREATE TABLE IF NOT EXISTS `audit` (
+CREATE TABLE IF NOT EXISTS `audit_records` (
   `id` BIGINT(20) NOT NULL AUTO_INCREMENT,
```
yogeshnachnani: do we need an id for this table?

db/create_tables.sql
```diff
@@ -1,4 +1,4 @@
...
   `state_machine_name` VARCHAR(255) NOT NULL,
```
yogeshnachnani: why have state_machine name and state machine instance id both?
and why are both varchars ?

db/create_tables.sql
```diff
@@ -1,4 +1,4 @@
...
   `state_machine_instance_id` VARCHAR(255) NOT NULL,
```
yogeshnachnani: should we have a created_at column over here?
Also, do we have one row per state of a state machine or 1 row for everything that happens to a state?

For eg: After successful execution of a state, will the table read
<sm_id,state_id,retryAttempt:1,state_status:Completed,..)
OR
<sm_id,state_id,retryAttempt:0,state_status:Failed,..)
<sm_id,state_id,retryAttempt:1,state_status:Completed)

db/create_tables.sql
```diff
@@ -21,7 +21,7 @@ CREATE TABLE IF NOT EXISTS `checkpoints` (
...
   `state_id` BIGINT(20) NOT NULL,
-  `data` LONGTEXT,
+  `data` BLOB,
```
yogeshnachnani: Lets not create this table in the first cut. We are not using it internally anywhere, correct?

model/src/main/java/com/flipkart/flux/domain/AuditRecord.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /** The State execution retry count */
+    private int retryAttempt;
+
+    /** The State execution status */
+    @Enumerated(EnumType.STRING)
+    private Status stateStatus;
+
+    /** Time when the State has started */
+    private Date stateStartTime;
```
yogeshnachnani: Use timestamp

model/src/main/java/com/flipkart/flux/domain/AuditRecord.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /** Time when the State has ended */
+    private Date stateEndTime;
```
yogeshnachnani: Use Timestamp

model/src/main/java/com/flipkart/flux/domain/AuditRecord.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /** Constructors */
+    public AuditRecord(){}
```
yogeshnachnani: package local

model/src/main/java/com/flipkart/flux/domain/AuditRecord.java
```diff
@@ -0,0 +1,111 @@
...
+        this.stateMachineName = stateMachineName;
+        this.stateMachineInstanceId = stateMachineInstanceId;
+        this.stateId = stateId;
+        this.retryAttempt = retryAttempt;
+        this.stateStatus = stateStatus;
+        this.stateStartTime = stateStartTime;
+        this.stateEndTime = stateEndTime;
+    }
+
+    /** Accessor/Mutator methods*/
```
yogeshnachnani: don't create getters/setters till you don't need them

yogeshnachnani: @regunathb - Or do you suggest anything different?

model/src/main/java/com/flipkart/flux/domain/Event.java
```diff
@@ -13,63 +13,123 @@
...
+import javax.persistence.*;
 import java.io.Serializable;
+import java.util.Date;
 
 /**
  * <code>Event</code> is the result of a {@link Task} execution.
  * It is to be posted back to the Flux execution engine once a worker has executed the task
- * 
+ *
  * @author Yogesh
```
yogeshnachnani: @regunathb - Do we really need these author tags considering everything is in git & this is open source?

model/src/main/java/com/flipkart/flux/domain/Event.java
```diff
@@ -13,63 +13,123 @@
...
+
+    /** Data associated with this Event, must have public getters and setters and be serializable */
+    @Type(type = "BlobType")
     private T eventData;
 
+    /** The source who generated this Event */
+    private String eventSource;
+
+    /** Event creation time */
+    @CreationTimestamp
```
yogeshnachnani: don't use hibernate to get the creationTimestamp. Let it come from the DB itself.
That way we won't have to bother about time differences between service nodes (if any)

model/src/main/java/com/flipkart/flux/domain/Event.java
```diff
@@ -13,63 +13,123 @@
...
+    }
 
-	/** Enum of Event statuses*/
+    /** Enum of Event statuses*/
     public enum EventStatus {
         pending,triggered;
     }
 
+    /** Constructors */
+    public Event() {}
```
yogeshnachnani: package local

model/src/main/java/com/flipkart/flux/domain/Event.java
```diff
@@ -13,63 +13,123 @@
...
+
+    /** Time at which this event is last updated */
+    @UpdateTimestamp
+    private Date updatedAt;
+
     /** Constructor*/
     public Event(String name, String type) {
-		super();
-		this.name = name;
-		this.type = type;
```
yogeshnachnani: Delete this constructor since you've created a complete one below

model/src/main/java/com/flipkart/flux/domain/State.java
```diff
@@ -46,27 +62,44 @@
...
     /** The Status of state transition execution*/
+    @Transient
     private Status status;
+
     /** The rollback status*/
+    @Transient
     private Status rollbackStatus;
+
     /** The number of retries attempted*/
+    @Transient
```
yogeshnachnani: Why are these transient?
You plan to source them from audit table, is it?

model/src/main/java/com/flipkart/flux/domain/StateMachine.java
```diff
@@ -13,73 +13,112 @@
...
     private Long version;
     /** Name for this state machine*/
     private String name;
     /** Description of the state machine*/
     private String description;
+
     /** List of states that this machine has*/
+    @OneToMany(fetch = FetchType.EAGER, cascade=CascadeType.ALL, targetEntity = State.class)
+    @JoinColumn(name = "stateMachineId")
     private List<State<T>> states;
```
yogeshnachnani: change it to Set<> states?

model/src/main/java/com/flipkart/flux/domain/StateMachine.java
```diff
@@ -13,73 +13,112 @@
...
+    @CreationTimestamp
+    private Date createdAt;
+
+    /** Time at which this State Machine has been last updated */
+    @UpdateTimestamp
+    private Date updatedAt;
+
+
+    /** Constructors*/
+    public StateMachine() {}
```
yogeshnachnani: package local

model/src/main/java/com/flipkart/flux/type/BlobType.java
```diff
@@ -0,0 +1,119 @@
...
+        } catch (IOException e) {
+            throw new SQLException("Cannot deserialize byte array. Exception " + e.getMessage());
+        } catch (ClassNotFoundException e) {
+            throw new SQLException("Cannot deserialize byte array. Exception " + e.getMessage());
+        }
+    }
+
+    @Override
+    public Object deepCopy(Object value) throws HibernateException {
+        return value;
```
yogeshnachnani: Why is deepCopy returning a reference?

model/src/main/java/com/flipkart/flux/type/BlobType.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Override
+    public boolean isMutable() {
+        return true;
```
yogeshnachnani: Shouldn't this be false?

model/src/main/java/com/flipkart/flux/type/StoreFQNType.java
```diff
@@ -0,0 +1,118 @@
...
+        } catch (InstantiationException e) {
+            throw new SQLException("Cannot build object of class: "+value+". Exception " + e.getMessage());
+        } catch (IllegalAccessException e) {
+            throw new SQLException("Cannot build object of class: "+value+". Exception " + e.getMessage());
+        }
+    }
+
+    @Override
+    public Object deepCopy(Object value) throws HibernateException {
+        return value;
```
yogeshnachnani: same here

model/src/main/java/com/flipkart/flux/type/StoreFQNType.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public boolean isMutable() {
+        return true;
```
yogeshnachnani: false?

pom.xml
```diff
@@ -31,10 +31,30 @@
...
             <version>5.1.38</version>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-core</artifactId>
+            <version>5.1.0.Final</version>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-entitymanager</artifactId>
```
yogeshnachnani: hibernate dependencies will go in model, right?

yogeshnachnani: my bad. They would go in runtime

runtime/src/main/java/com/flipkart/flux/dao/AbstractDAO.java
```diff
@@ -0,0 +1,89 @@
...
+import org.hibernate.criterion.Restrictions;
+
+/**
+ * @author shyam.akirala
+ */
+public class AbstractDAO<T> {
+
+    public T findById(Class cls, Long id) {
+        Session session = HibernateUtil.getSessionFactory().getCurrentSession();
+        Transaction tx = session.beginTransaction();
```
yogeshnachnani: Why a transaction in a read only method?

runtime/src/main/java/com/flipkart/flux/dao/AbstractDAO.java
```diff
@@ -0,0 +1,89 @@
...
+import com.flipkart.flux.util.HibernateUtil;
+import org.hibernate.Criteria;
+import org.hibernate.Session;
+import org.hibernate.Transaction;
+import org.hibernate.criterion.Restrictions;
+
+/**
+ * @author shyam.akirala
+ */
+public class AbstractDAO<T> {
```
yogeshnachnani: Lets not open sessions or transactions in the dao.
Lets do it using AOP/Decorators (similar to DW UnitOfWork). That way we can keep it at any layer above the DAO layers (or on the DAO layers) that may need to work with multiple DAOs and thus need a session and tx across the whole shebang

runtime/src/main/java/com/flipkart/flux/dao/AuditDAOImpl.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.util.List;
+
+/**
+ * @author shyam.akirala
+ */
+public class AuditDAOImpl extends AbstractDAO<AuditRecord> implements AuditDAO {
+
+    @Override
+    public AuditRecord findById(Long id) {
```
yogeshnachnani: You can take the type of the class in the constructor of the AbstractDAO and thus you will not need to unnecessarily implement these methods that do nothing but call the super() with the class reference

runtime/src/main/java/com/flipkart/flux/dao/AbstractDAO.java
```diff
@@ -0,0 +1,89 @@
...
+        Session session = HibernateUtil.getSessionFactory().getCurrentSession();
+        Transaction tx = session.beginTransaction();
+
+        session.save(object);
+
+        tx.commit();
+        return object;
+    }
+
+    public void update(T object) {
```
yogeshnachnani: aren't these implemented in AbstractDao?

runtime/src/main/java/com/flipkart/flux/dao/AuditDAOImpl.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+
+    @Override
+    public Long create(AuditRecord auditRecord) {
+        return super.save(auditRecord).getId();
+    }
+
+    @Override
+    public void update(AuditRecord auditRecord) {
```
yogeshnachnani: ?

runtime/src/main/java/com/flipkart/flux/dao/iface/AuditDAO.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.flipkart.flux.domain.AuditRecord;
+import java.util.List;
+
+/**
+ * @author shyam.akirala
+ */
+public interface AuditDAO {
+
+    AuditRecord findById(Long id);
```
yogeshnachnani: when will we need this?

runtime/src/main/java/com/flipkart/flux/dao/iface/AuditDAO.java
```diff
@@ -0,0 +1,34 @@
...
+
+    List<AuditRecord> findBySMInstanceId(String stateMachineInstanceId);
+
+    AuditRecord find(String stateMachineInstanceId, Long stateId, int retryAttempt);
```
yogeshnachnani: when will we need this?

runtime/src/main/java/com/flipkart/flux/util/HibernateUtil.java
```diff
@@ -0,0 +1,49 @@
...
+package com.flipkart.flux.util;
+
+import com.flipkart.flux.domain.*;
+import org.hibernate.SessionFactory;
+import org.hibernate.cfg.Configuration;
+
+/**
+ * @author shyam.akirala
+ */
+public class HibernateUtil {
```
yogeshnachnani: Why is all this in a static block?

runtime/src/main/java/com/flipkart/flux/util/HibernateUtil.java
```diff
@@ -0,0 +1,49 @@
...
+    private static final SessionFactory sessionFactory;
+
+    static {
+        try {
+            Configuration  configuration = new Configuration().configure("hibernate.cfg.xml");
+            addAnnotatedClasses(configuration);
+            configuration.setPhysicalNamingStrategy(PhysicalNamingStrategyImpl.INSTANCE);
```
yogeshnachnani: Why not use the standard hibernate conventions?

runtime/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
yogeshnachnani: I suggest we use the log4j.properties file. Its much more readable. 
@regunathb - makes sense?

runtime/src/test/java/com/flipkart/flux/dao/AuditDAOTest.java
```diff
@@ -0,0 +1,52 @@
...
+public class AuditDAOTest {
+
+    @Test
+    public void createAuditRecordTest() {
+        AuditDAO auditDAO = new AuditDAOImpl();
+        AuditRecord auditRecord = new AuditRecord("test_state_machine_name", "test_state_machine_instance_id", 10L, 0, Status.initialized, new Date(), null);
+        auditDAO.create(auditRecord);
+
+        List<AuditRecord> records = auditDAO.findBySMInstanceId("test_state_machine_instance_id");
+        Assert.assertNotNull(records);
```
yogeshnachnani: so if I change the create impl to create 10 entries instead of 1, this test would still pass :)

runtime/src/test/java/com/flipkart/flux/dao/DummyOnEntryHook.java
```diff
@@ -0,0 +1,29 @@
...
+
+package com.flipkart.flux.dao;
+
+import com.flipkart.flux.domain.Event;
+import com.flipkart.flux.domain.Hook;
+
+/**
+ * @author shyam.akirala
+ */
+public class DummyOnEntryHook implements Hook {
```
yogeshnachnani: why not use mocks?

runtime/src/test/java/com/flipkart/flux/dao/EventsDAOTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Test
+    public void createEventTest() {
+        EventsDAO eventsDAO = new EventsDAOImpl();
+        EventData data = new EventData("external_event", "test_event_info");
+        Event<EventData> event = new Event("test_name","test_type", Event.EventStatus.pending,"test_state_machine_instance_id", data,"internal_event");
+        eventsDAO.create(event);
+
+        List<Event> events = eventsDAO.findBySMInstanceId("test_state_machine_instance_id");
+        Assert.assertNotNull(events);
```
yogeshnachnani: same here :) Assert the actual event

yogeshnachnani:
I didn't see where we are using liquibase. Also, how do we plan to run migrations in stage/prod envs?

db/create_tables.sql
shyamakirala: I've committed DB schema to master earlier itself, before starting DAO code.

runtime/src/main/java/com/flipkart/flux/util/HibernateUtil.java
shyamakirala: Standard hibernate/jpa conventions suggests to use Camel case for table and column names. Thought using underscore notation would be more readable, apart from that there is no reason to avoid Camel case, we can go with Camel case if you guys want.

db/create_tables.sql
shyamakirala: Right now I don't see any use of Id in audit table, will remove it and keep state_machine_instance_id, state_id and retry number combination as key.

db/create_tables.sql
shyamakirala: We can save SM execution status and related context in this table for replay purposes.

db/create_tables.sql
shyamakirala: Thinking to use UUID to identify state machine instance, that's why kept it as varchar.
Have put state machine name to maintain mapping between state machine and state machine instance id for offline query purposes, will remove it from here as it is available in checkpoint table.

db/create_tables.sql
shyamakirala: we can have created_at.
It is the second one.
<sm_id,state_id,retryAttempt:0,state_status:Failed,..)
<sm_id,state_id,retryAttempt:1,state_status:Completed)

db/create_tables.sql
shyamakirala: same as above. Right now I don't see any use of id column. We can remove it.

shyamakirala: my bad, we need it. When the event is fired it would in pending state, and once it is consumed successfully it would go triggered status. That's what I inferred from Event class status column. During update we can use event id to identify the particular event.

db/create_tables.sql
shyamakirala: same as above. We may need it, if are planning to change the status of the event.

db/create_tables.sql
shyamakirala: state_machine table doesn't contain state machine instance details. It is to store only state metadata.

db/create_tables.sql
shyamakirala: TINYINT is capable of storing up to 255. That's a big number for version column IMO considering each version would be having multiple SMs with unique internal Ids.
If you think it may cross, then we can go for next class SMALL INT with number till 65535.

db/create_tables.sql
shyamakirala: We can ignore this if we are generating a DAG for SM, and start execution of all independent states.

The state machine class has start state variable, so thought we are going with it.

db/create_tables.sql
shyamakirala: will change all created and updated timestamps to get updated from DB side.

db/create_tables.sql
shyamakirala: It keeps in this format "2015-08-31 15:22:25.535". I think having 3 digits for fractional seconds is good enough.

db/create_tables.sql
shyamakirala: The names of events the state depends on.

db/create_tables.sql
shyamakirala: same as above. IMO numbers till 255 are enough for retry_count.

model/src/main/java/com/flipkart/flux/domain/AuditRecord.java
shyamakirala: It stores in format "2015-08-31 15:22:25.535". Will compare it with timestamp and see if it offers any additional things.

model/src/main/java/com/flipkart/flux/domain/AuditRecord.java
shyamakirala: It stores in format "2015-08-31 15:22:25.535". Will compare it with timestamp and see if it offers any additional things.

model/src/main/java/com/flipkart/flux/domain/AuditRecord.java
shyamakirala: will change it.

model/src/main/java/com/flipkart/flux/domain/Event.java
shyamakirala: will change it to put timestamps at db layer.

model/src/main/java/com/flipkart/flux/domain/State.java
shyamakirala: State table will store only meta data, that's why I kept these as transient as these are only runtime fields.
Yes, we can take these from Audit table. (I missed rollback status in audit table, will add it)

model/src/main/java/com/flipkart/flux/type/BlobType.java
shyamakirala: Didn't put much thought into this. Took the implementation from Marvin :)

model/src/main/java/com/flipkart/flux/type/BlobType.java
shyamakirala: The data object can be changed.

model/src/main/java/com/flipkart/flux/type/StoreFQNType.java
shyamakirala: will change it to false.

runtime/src/main/java/com/flipkart/flux/dao/AbstractDAO.java
shyamakirala: Criteria object can't be used outside transaction boundaries, so had to start transaction.

runtime/src/main/java/com/flipkart/flux/dao/AuditDAOImpl.java
shyamakirala: Thought of keeping state start time and end time one place for a particular retry, but it will clash with status column. So it's good to put multiple records for status changes as well. Will remove it.

runtime/src/main/java/com/flipkart/flux/util/HibernateUtil.java
shyamakirala: Keeping a single SessionFactory instance and intializing it when class loads up, that's why put it in static block.

runtime/src/test/java/com/flipkart/flux/dao/DummyOnEntryHook.java
shyamakirala: I wanted to store the proper fqn of class in db, and retrieve back and load the class using reflection. Not sure using mocks I can do this.

runtime/src/main/resources/log4j2.xml
shyamakirala: log4j2 has many performance improvements compared to it's previous version.
http://www.infoq.com/news/2014/07/apache-log4j2

.properties file support has been removed in the new version. It supports only xml, json and yaml files.

db/create_tables.sql
yogeshnachnani: try to raise the whole feature as a pull request from next time

db/create_tables.sql
yogeshnachnani: lets remove it.

db/create_tables.sql
yogeshnachnani: lets go with int.

db/create_tables.sql
yogeshnachnani: Does it support automatic update of the column values?

db/create_tables.sql
yogeshnachnani: sure

db/create_tables.sql
yogeshnachnani: whats a status of an event?

db/create_tables.sql
yogeshnachnani: I really didn't understand.
If we are linking two tables, why not use the PK of one as FK in the other?

db/create_tables.sql
yogeshnachnani: got it

db/create_tables.sql
yogeshnachnani: No, that data can be derived out of Events & State & State Machine table anyway, right?

model/src/main/java/com/flipkart/flux/type/BlobType.java
yogeshnachnani: Thats wrong and misleading. Please fix it

model/src/main/java/com/flipkart/flux/type/BlobType.java
yogeshnachnani: No it can't. A data object returned by a task is immutable

runtime/src/main/java/com/flipkart/flux/dao/AbstractDAO.java
yogeshnachnani: Really? I thought all that was required is a session

runtime/src/main/java/com/flipkart/flux/util/HibernateUtil.java
yogeshnachnani: lets stick to conventions

runtime/src/main/resources/log4j2.xml
yogeshnachnani: didn't know that! thanks for the info :)

runtime/src/test/java/com/flipkart/flux/dao/DummyOnEntryHook.java
yogeshnachnani: Sure you can

yogeshnachnani:
please make the necessary changes and let us know so we can re-review it

db/create_tables.sql
yogeshnachnani: lets discuss this f2f

runtime/src/main/java/com/flipkart/flux/util/HibernateUtil.java
yogeshnachnani: If you want a singleton, you can provide it using whatever DI tool we use. That way we get a complete hold on its lifecycle - which we won't if it is static.

PR #4, opened 2016-04-11
Round 1 self code review changes in module - model

kartikaditya:
1. Added hashcode/equals which ever was used in Set<>
2. Made StateMachine has multiple current states and not single
3. Made Map or Map for DAG in Context
4. Added version in EventDefinition

### DIFF
--- a/model/src/main/java/com/flipkart/flux/domain/Context.java
+++ b/model/src/main/java/com/flipkart/flux/domain/Context.java
@@ -24,6 +24,7 @@ import java.util.Map;
  * @author Yogesh
  * @author regunath.balasubramanian
  * @author shyam.akirala
+ * @author kartik.bommepally
  */
 public abstract class Context<T> {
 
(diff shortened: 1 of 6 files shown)


kartikaditya:
@yogeshnachnani - Please review. I am unable to edit Assignee.

model/src/main/java/com/flipkart/flux/domain/Context.java
```diff
@@ -32,7 +33,7 @@
     /** Identifier for the Context*/
     protected String contextId;
     /** A dependency graph created across States - holds information on possible next state transitions for a State, keyed by the Event FQN*/
-    protected Map<String,List<State<T>>> stateToEventDependencyGraph;
+    protected Map<String, Map<String, State<T>>> stateToEventDependencyGraph;
```
yogeshnachnani: Why a map of maps?

kartikaditya: My understanding is stateToEventDependencyGraph must represent DAG for the StateMachine.
Now, from a State its Task can raise *several* Events. And each Event will land up on another *single* State.
So its a Map(PivotState, Map(EventFqn, TransitionState))

yogeshnachnani: No, its a reverse dependency graph. 
It means that a given event (identified by string for now) has the following list of states waiting on it.
Once we get the event, we can iterate over the list and figure if any of the states is unblocked.
We can enter all unblocked states at once

Also, a task will raise only event. The return value of the method that is executed as per the task execution is wrapped in an event and sent back to the runtime

kartikaditya: Ah ok, will revert back and improve the doc.

PR #5, opened 2016-04-13
skeleton for the workflow and task interceptor framework

yogeshnachnani:
This is unfinished.
Keeping it open for early comments/questions


client/pom.xml
```diff
@@ -42,6 +42,35 @@
             <artifactId>javax.inject</artifactId>
             <version>${javax.inject.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.inject</groupId>
+            <artifactId>guice</artifactId>
+            <version>${guice.version}</version>
+        </dependency>
+        <dependency>
```
kartikaditya: Can we move such deps to parent level?

yogeshnachnani: Since the client jar will actually be imported by the user, I wanted to keep the dependencies with the parent pom to an absolute minimum.
I was thinking of actually using the actual version number rather than the variable :)

client/src/main/java/com/flipkart/flux/client/runtime/LocalContext.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ */
+
+package com.flipkart.flux.client.runtime;
+
+/**
+ * Maintains all local flux related context
+ * @author yogesh.nachnani
+ */
+public class LocalContext {
```
kartikaditya: Not here.
But can we rename the existing com.flipkart.flux.domain.Context to ExecutionContext or a better name to avoid confusion?

yogeshnachnani:
@kartikaditya @ashishbhutani @shyamakirala - Please have a look

api/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+    <artifactId>flux</artifactId>
+    <version>1.0-SNAPSHOT</version>
+  </parent>
+  <groupId>com.flipkart.flux</groupId>
+  <artifactId>api</artifactId>
+  <version>1.0-SNAPSHOT</version>
+  <name>api</name>
+  <url>http://maven.apache.org</url>
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
shyamakirala: This can be removed, as it's in the root pom.

api/src/test/java/com/flipkart/flux/api/WorkflowStateSummaryTest.java
```diff
@@ -10,24 +10,33 @@
...
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.commons.io.IOUtils;
+import org.hamcrest.core.Is;
+import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 
 import static org.hamcrest.core.Is.is;
 import static org.junit.Assert.assertThat;
```
shyamakirala: optimize imports

api/src/test/java/com/flipkart/flux/api/WorkflowStatesDetailTest.java
```diff
@@ -10,23 +10,34 @@
...
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.commons.io.IOUtils;
+import org.hamcrest.CoreMatchers;
+import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertThat;
```
shyamakirala: optimize imports

client/src/main/java/com/flipkart/flux/client/intercept/WorkflowInterceptor.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Inject
+    private LocalContext localContext;
+    @Inject
+    private FluxRuntimeConnector connector;
+
+    public WorkflowInterceptor() {
+    }
+
+    public WorkflowInterceptor(LocalContext localContext, FluxRuntimeConnector connector) {
```
shyamakirala: This constructor can be removed, as this will not be used anywhere.

client/src/main/java/com/flipkart/flux/client/intercept/TaskInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+public class TaskInterceptor implements MethodInterceptor {
+
+    @Inject
+    private LocalContext localContext;
+
+    public TaskInterceptor() {
+    }
+
+    TaskInterceptor(LocalContext localContext) {
```
shyamakirala: This constructor can be removed, as not in use.

client/src/main/java/com/flipkart/flux/client/intercept/TaskInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+        this.localContext = localContext;
+    }
+
+    @Override
+    public Object invoke(MethodInvocation invocation) throws Throwable {
+        final Method method = invocation.getMethod();
+        final Task taskAnnotation = method.getAnnotationsByType(Task.class)[0];
+        localContext.registerNewState(taskAnnotation.version(), generateStateIdentifier(method) ,null,null, generateTaskIdentifier(method),taskAnnotation.retries(),taskAnnotation.timeout(),generateEventDefs(method));
```
shyamakirala: 1. Here we are constructing task identifiers with underscore notation, but state machine definition expects FQN of task class, this needs to be modified.
2. Hook class name needs be passed instead of null.
3. The same hook parameter is getting used as onEntry and onExit hooks in registerNewState method implementation, which also need to changed. (I'm fine with having one hook class with preEntry and postExecute methods, will discuss f2f)

client/src/main/java/com/flipkart/flux/client/intercept/WorkflowInterceptor.java
yogeshnachnani: its used in tests

client/src/main/java/com/flipkart/flux/client/intercept/TaskInterceptor.java
yogeshnachnani: used in tests

client/src/main/java/com/flipkart/flux/client/intercept/TaskInterceptor.java
yogeshnachnani: 1. StateMachineDefinition works out of string - it can be any string. Underscore is used to separate different types (see tests and you'll know what i'm talking about)
2. Yes, hook support will be added later
3. We'll figure that out when we add hook support. Its not that hard

api/src/test/java/com/flipkart/flux/api/WorkflowStateSummaryTest.java
yogeshnachnani: I'll do it if you can tell me why I should :)

shyamakirala: bcoz it's a good practise :D 
apart from that optimizing in test classes won't do much, as those would be local.

client/src/main/java/com/flipkart/flux/client/intercept/TaskInterceptor.java
shyamakirala: 1. got it. it can be extracted anyway.
2 & 3. sure.

api/src/test/java/com/flipkart/flux/api/WorkflowStateSummaryTest.java
yogeshnachnani: 1. Why do you say it is good practice ? :)
2. It doesn't have any performance impact in any class - even if its a production class

yogeshnachnani: I optimised and merged since the changes were needed in master. But lets continue this discussion no :)

shyamakirala: In my view optimizing imports improves the code readability. That's why I always prefer to do it (lot of people prefer the same).
In this case it won't apply, but cases like importing .* creates unnecessary linkages at compilation time (If I remember correctly), It doesn't impact runtime performance though.

yogeshnachnani: 1. How does it improve readability ? Most IDEs hide all import statements in one single line any way. 
2. It doesn't create any linkages 

Please don't think i'm dragging on for the heck of it. It is important that we bring this to a logical conclusion

shyamakirala: :) it's fine. Have a look at
http://www.javaperformancetuning.com/news/qotm031.shtml

yogeshnachnani: sure, thats one of the first results google throws :p 
How does it impact readability with modern IDEs? :)

PR #6, opened 2016-04-15
Supporting yml flat properties binding/injection

kartikaditya:
1. Added ConfigModule to bind
1.1 Flattened view of the yml properties
1.2 Polyguices - YamlConfiguration & ApacheCommonsConfigProvider
2. With this we can support usecases such as in ContainerModule which has @Named("dashboard.service.port") int port
3. Cleaned up HibernateModule bindings. Removed explicit Providers and added @Provides method instead

@shyamakirala @regunathb - Please review


shyamakirala:
I already did these changes and committed to my branch :) That's fine. Will review this.

kartikaditya:
Ah duplication counter: 1

runtime/src/main/java/com/flipkart/flux/guice/module/HibernateModule.java
```diff
@@ -55,4 +65,49 @@ protected void configure() {
...
+     * Since the yaml properties are already flattened in input param <code>yamlConfiguration</code>
+     * the method loops over them to selectively pick Hibernate specific properties.
+     */
+    @Provides
+    @Singleton
+    public Configuration getConfiguration(YamlConfiguration yamlConfiguration) {
+        Configuration configuration = new Configuration();
+        addAnnotatedClassesAndTypes(configuration);
+        Iterator<String> propertyKeys = yamlConfiguration.getKeys();
+        while (propertyKeys.hasNext()) {
```
shyamakirala: Instead of checking startsWith you can use method the below method.
Configuration configuration = yamlConfiguration.subset(key); //key is Hibernate in this case
This returns all config properties which starts with "Hibernate."(including dot), basically which are under element Hibernate

runtime/src/main/java/com/flipkart/flux/guice/module/ConfigModule.java
```diff
@@ -0,0 +1,63 @@
...
+import com.google.inject.binder.LinkedBindingBuilder;
+import com.google.inject.name.Names;
+
+/**
+ * <code>ConfigModule</code> is a Guice {@link AbstractModule} implementation used for wiring flux configuration.
+ * @author kartik.bommepally
+ */
+public class ConfigModule extends AbstractModule {
+
+    private static final String CONFIGURATION_YML = "packaged/configuration.yml";
```
shyamakirala: Keep this in RutimeConstants file, in this way we can have all variables in one place.

runtime/src/main/java/com/flipkart/flux/guice/module/HibernateModule.java
kartikaditya: Thanks for the suggestion. Will do

runtime/src/main/java/com/flipkart/flux/guice/module/ConfigModule.java
kartikaditya: Done

PR #7, opened 2016-04-16
Bringing up dashboard webapp

kartikaditya:
1. Fixed some code in the ControllerModule to bring up the webapp
2. Added relevant config to yml file
3. Added simple test case to bring up the service

### DIFF
--- a/runtime/src/main/java/com/flipkart/flux/config/FileLocator.java
+++ b/runtime/src/main/java/com/flipkart/flux/config/FileLocator.java
@@ -15,13 +15,13 @@
  */
 package com.flipkart.flux.config;
 
+import static com.flipkart.flux.constant.RuntimeConstants.CONFIG_ROOT;
+
 import java.io.File;
 import java.io.IOException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Enumeration;
-
-import javax.inject.Inject;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -33,7 +33,8 @@ import com.flipkart.flux.constant.RuntimeConstants;
 * The <code>FileLocator</code> is a utility for locating configuration files deployed under the projects root. This class provides methods to locate
 * all occurrences of a file with the specified name. This class looks into specific directories under the project root folder to locate files.
 *  
-* @author regunath.balasubramanian 
+* @author regunath.balasubramanian
+* @author kartik.bommepally 
 */
 public class FileLocator {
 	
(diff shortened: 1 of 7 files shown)


kartikaditya:
@yogeshnachnani @shyamakirala - Kindly review

runtime/src/main/java/com/flipkart/flux/config/FileLocator.java
```diff
@@ -42,9 +43,11 @@
 	 */
 	private static final Logger LOGGER = LogManager.getLogger(FileLocator.class);
 	
-	/** The root folder where Flux runtime is deployed*/
-	@Inject
-	private static String deploymentRoot;
+	/** The config root folder of Flux */
+	private static final String DEPLOYED_CONFIG_ROOT;
```
yogeshnachnani: Static?
We wanted it to be a configuration, hence added the Inject annotation

runtime/src/test/java/com/flipkart/flux/controller/DashboardWebAppTest.java
```diff
@@ -0,0 +1,95 @@
...
+ * @author kartik.bommepally
+ */
+public class DashboardWebAppTest {
+
+    private static Injector injector;
+    private static Server server;
+    private static int port;
+
+    @BeforeClass
+    public static void beforeClass() throws Exception {
```
yogeshnachnani: Could you please leave a TODO for me to replace this with the guiceJunit4Runner? I have added it in my branch, will pull it into this one as well

yogeshnachnani:
looks good - we can change everything from RunTimeConstants into configuration whenever we want.

yogeshnachnani:
i'll leave it for the assignee to merge and close this :)

runtime/src/main/java/com/flipkart/flux/config/FileLocator.java
kartikaditya: This was my confusion too, when I read the code.
We cannot inject a static var, so should I be making deploymentRoot and thus FileLocator injectable?

runtime/src/test/java/com/flipkart/flux/controller/DashboardWebAppTest.java
kartikaditya: Sure, will do

runtime/src/main/java/com/flipkart/flux/config/FileLocator.java
yogeshnachnani: sure we can - https://github.com/google/guice/wiki/Injections

However, I agree we shouldn't. We can replace this with a decent builder or to constructor based injection. I don't see it as a pressing issue, but if you feel strongly about it then you can go ahead and do that - it'll be a trivial change

kartikaditya: I'l add a TODO to make it config based for now

PR #10, opened 2016-05-12
Event orchestration changes

shyamakirala:


runtime/src/main/java/com/flipkart/flux/dao/StateMachinesDAOImpl.java
```diff
@@ -47,20 +50,20 @@ public StateMachine findById(Long id) {
 
     @Override
     @Transactional
-    public List<StateMachine> findByName(String stateMachineName) {
+    public Set<StateMachine> findByName(String stateMachineName) {
```
shyamakirala: Changed it to Set, as Hibernate is returning duplicate objects of same state machine, in case number of states are more than 1 (due to state machine and state tables join). Currently states fetch type is Eager, if we change it to Lazy, this issue can be avoided, but I don't want to change it, as if we try to access states outside transaction boundaries, we get "no session bound currently" exception (only in tests we are accessing states after session close right now).

api/src/main/java/com/flipkart/flux/api/EventData.java
```diff
@@ -0,0 +1,72 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.flipkart.flux.api;
+
+/**
+ * <code>EventData</code> represents the event which would be submitted to flux runtime from Akka system.
```
yogeshnachnani: from any system, not just akka

model/src/main/java/com/flipkart/flux/domain/Context.java
```diff
@@ -52,15 +54,45 @@
...
 
+    /**
+     * This builds dependency graph between event and states and keeps for later use. Currently dependency graph is created on every event arrival.
+     */
+    public void buildDependencyMap(Set<State> states) {
+        stateToEventDependencyGraph = new HashMap<>();
+        for(State state : states) {
+            if (state.getDependencies() != null) {
+                for (String eventName : state.getDependencies()) {
+                    if (stateToEventDependencyGraph.containsKey(eventName)) {
```
yogeshnachnani: its easier if you invert this. if (!containsKey()) then add key with empty set.
Then, you always get() the set by the key and add the new member

model/src/main/java/com/flipkart/flux/domain/Context.java
```diff
@@ -52,15 +54,45 @@
...
         return false;
     }
 
+    /**
+     * This builds dependency graph between event and states and keeps for later use. Currently dependency graph is created on every event arrival.
+     */
+    public void buildDependencyMap(Set<State> states) {
+        stateToEventDependencyGraph = new HashMap<>();
+        for(State state : states) {
+            if (state.getDependencies() != null) {
```
yogeshnachnani: a returned set should never be null, it can be empty but never null.
With that assumption, this check will change to an empty check

model/src/main/java/com/flipkart/flux/domain/Context.java
```diff
@@ -52,15 +54,45 @@
...
+                    } else {
+                        Set set = new HashSet<State>() {{
+                            add(state);
+                        }};
+                        stateToEventDependencyGraph.put(eventName, set);
+                    }
+                }
+            } else {
+                if (stateToEventDependencyGraph.containsKey(null)) {
+                    stateToEventDependencyGraph.get(null).add(state);
```
yogeshnachnani: it'll be better if we use a constant instead of a null. say a "." ?

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -13,32 +13,122 @@
...
-     * Trigger state machine execution.
+     * Perform init operations on a state machine and starts execution of states which are not dependant on any events.
      * @param stateMachine
      * @return List of states that do not have any event dependencies on them
      */
-    List<State<T>> init(StateMachine<T> stateMachine);
+    public Set<State> initAndStart(StateMachine stateMachine) {
+        Context context = new RAMContext(System.currentTimeMillis(), null); //TODO: set context id, should we need it ?
+        stateMachine.setContext(context);
+        context.buildDependencyMap(stateMachine.getStates());
```
yogeshnachnani: why a separate call to buildDependencyMap and not construct the same map as part of object construction?

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -13,32 +13,122 @@
...
+
+        Set<State> dependantStates = context.getDependantStates(null); //passing dependencies as null, start all the states which are not dependant on any events.
```
yogeshnachnani: instead of context.getDependentStates(null) can we have context.getInitialStates() and in turn it is looking up the hashmap for the "null" or the "." placeholder? That way knowledge of what context does internally is not known by other classes

shyamakirala: done.

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
shyamakirala: done. As context and SM have one-to-one relation, attaching the context to state machine and building dependency graph as part of construction itself.

model/src/main/java/com/flipkart/flux/domain/Context.java
shyamakirala: done.

model/src/main/java/com/flipkart/flux/domain/Context.java
shyamakirala: @yogeshnachnani I think it would be better if we keep event name as "start" if the state is not depending on any event.  In dependency graph also we can use the same.

yogeshnachnani:
could you push those changes please?

model/src/main/java/com/flipkart/flux/domain/Context.java
yogeshnachnani: anything is fine - as long as it is a constant field in the Context class, the value doesn't matter

PR #11, opened 2016-05-25
small change for akka actor supervise strategy and UT for akka actor functionality

ashishbhutani:
*Default strategy of Akka in case of exception is to escalate and leads to re-start of all actors (remote as well). Changed that to use a custom one where based on exception type we can take decision to resume/restart/stop actors
*UT for functionality test of Akka actor. This just checks the functionality and isn't async, which the actual actor will be. Integration test shall check the async behavior.


yogeshnachnani:
can you please merge it with the latest master so i can review in one shot ?

ashishbhutani:
merged to master

yogeshnachnani:
I meant merge master branch into it_happens and push in it_happens so that I can review in one shot (since there were merge conflicts). My bad, should have been more specific

PR #12, opened 2016-05-25
Deployment unit changes

shyamakirala:
- create url class loader per deployment unit.


yogeshnachnani:
@shyamakirala - shall i start reviewing?

shyamakirala:
@yogeshnachnani Let me fix the UTs, after that you can start, will let you know

client/src/main/java/com/flipkart/flux/client/intercept/TaskInterceptor.java
```diff
@@ -162,7 +162,7 @@ private String generateStateIdentifier(Method method) {
         return method.getName();
     }
 
-    private String generateTaskIdentifier(Method method) {
+    public String generateTaskIdentifier(Method method) {
```
yogeshnachnani: use new MethodId(method).toString() wherever you need a task identifier.
Make this private again. 
You have included this class in another class just for this method :) Keep the boundaries clean

shyamakirala: biggest blunder :) my bad, thought we were calling some method which has the implementation in task interceptor itself, and added line note also saying we should move it to some util class. Ignore it.

examples/src/main/java/com/flipkart/flux/examples/concurrent/Email.java
```diff
@@ -35,6 +35,22 @@ public Email(String body, String recipient) {
         this.recipient = recipient;
     }
 
+    public String getBody() {
```
yogeshnachnani: Is this still required? I thought JsonProperty annotations now started working

examples/run_example.sh
```diff
@@ -1,7 +1,25 @@
...
+mkdir -p $DEPLOYMENT_UNIT_PATH/$DEPLOYMENT_UNIT_NAME/main
+mkdir -p $DEPLOYMENT_UNIT_PATH/$DEPLOYMENT_UNIT_NAME/lib
+
+echo "Copying jars to deployment unit"
+cp target/examples-1.0-SNAPSHOT.jar $DEPLOYMENT_UNIT_PATH/$DEPLOYMENT_UNIT_NAME/main
+cp target/dependency/* $DEPLOYMENT_UNIT_PATH/$DEPLOYMENT_UNIT_NAME/lib
+cp src/main/resources/flux_config.yml $DEPLOYMENT_UNIT_PATH/$DEPLOYMENT_UNIT_NAME/
+
+echo "Starting flux runtime"
+java -cp "target/dependency/*" "com.flipkart.flux.initializer.FluxInitializer" &
```
yogeshnachnani: Anyway we can color code this process's stdout to be different from the next (example) process's stdout?

examples/src/main/java/com/flipkart/flux/examples/concurrent/Email.java
yogeshnachnani: I mean it for all Event type classes used across the examples

runtime/pom.xml
```diff
@@ -71,6 +71,11 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.flipkart.flux</groupId>
```
yogeshnachnani: ?

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
```diff
@@ -0,0 +1,212 @@
...
+    private static final String CONFIG_FILE = "flux_config.yml";
+
+    /** Key in the config file, which has list of workflow/task class FQNs*/
+    private static final String WORKFLOW_CLASSES = "workflowClasses";
+
+    private static final String SLASH = "/";
+
+    /** Location of deployment units on the system*/
+    @Inject(optional=true) @Named("deploymentUnitsPath")
+    private String DEPLOYMENT_UNITS_PATH;
```
yogeshnachnani: only constants should have CAPS names

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -125,8 +125,15 @@ private void executeStates(Long stateMachineInstanceId, Set<State> executableSta
                 stateMachineInstanceId,
                 state.getOutputEvent(), state.getRetryCount());
             logger.debug("Sending msg {} for state machine {}", msg, stateMachineInstanceId);
-            this.routerRegistry.getRouter("someRouter").tell(
-                msg, ActorRef.noSender());
+            String taskName = state.getTask();
+            int secondUnderscorePosition = taskName.indexOf('_', taskName.indexOf('_') + 1);
+            String routerName = taskName.substring(0, secondUnderscorePosition == -1 ? taskName.length() : secondUnderscorePosition); //the name of router would be classFQN_taskMethodName
+            ActorRef router = this.routerRegistry.getRouter(routerName);
+            if(router == null) { //this should true only for E2E Tests
```
yogeshnachnani: we're planning to change this soon, I hope

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -125,8 +125,15 @@ private void executeStates(Long stateMachineInstanceId, Set<State> executableSta
                 stateMachineInstanceId,
                 state.getOutputEvent(), state.getRetryCount());
             logger.debug("Sending msg {} for state machine {}", msg, stateMachineInstanceId);
-            this.routerRegistry.getRouter("someRouter").tell(
-                msg, ActorRef.noSender());
+            String taskName = state.getTask();
+            int secondUnderscorePosition = taskName.indexOf('_', taskName.indexOf('_') + 1);
+            String routerName = taskName.substring(0, secondUnderscorePosition == -1 ? taskName.length() : secondUnderscorePosition); //the name of router would be classFQN_taskMethodName
```
yogeshnachnani: use methodId.getPrefix()

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /**
+     * Given a class loader, retrieves workflow classes names from config file, and returns methods
+     * which are annotated with {@link com.flipkart.flux.client.model.Workflow} annotation in those classes.
+     * @param urlClassLoader
+     * @return set of Methods
+     * @throws ClassNotFoundException
+     */
+    public Set<Method> getWorkflowMethods(URLClassLoader urlClassLoader) throws ClassNotFoundException, IOException {
```
yogeshnachnani: where is this method called ?

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
```diff
@@ -0,0 +1,212 @@
...
+     * Given a class loader, retrieves workflow classes names from config file, and returns methods
+     * which are annotated with {@link com.flipkart.flux.client.model.Task} annotation in those classes.
+     * @param urlClassLoader
+     * @return set of Methods
+     * @throws ClassNotFoundException
+     */
+    public Set<Method> getTaskMethods(URLClassLoader urlClassLoader) throws ClassNotFoundException, IOException {
+
+        YamlConfiguration yamlConfiguration = getProperties(urlClassLoader);
+        List<String> classNames = (List<String>) yamlConfiguration.getProperty(WORKFLOW_CLASSES);
```
yogeshnachnani: how come getTaskMethods is using WORKFLOW_CLASSES ?

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
```diff
@@ -0,0 +1,212 @@
...
+        Set<Method> methods = new HashSet<>();
+
+        //loading this class separately in this class loader as the following isAnnotationPresent check returns false, if
+        //we use default class loader's Task, as both class loaders don't have any relation between them.
+        Class taskClass = urlClassLoader.loadClass(Task.class.getCanonicalName());
```
yogeshnachnani: s/tasskClass/taskAnnotationClass ?

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
```diff
@@ -0,0 +1,212 @@
...
+     */
+    public Set<Method> getWorkflowMethods(URLClassLoader urlClassLoader) throws ClassNotFoundException, IOException {
+
+        YamlConfiguration yamlConfiguration = getProperties(urlClassLoader);
+        List<String> classNames = (List<String>) yamlConfiguration.getProperty(WORKFLOW_CLASSES);
+        Set<Method> methods = new HashSet<>();
+
+        //loading this class separately in this class loader as the following isAnnotationPresent check returns false, if
+        //we use default class loader's Workflow, as both class loaders don't have any relation between them.
+        Class workflowClass = urlClassLoader.loadClass(Workflow.class.getCanonicalName());
```
yogeshnachnani: s/workflowClass/workflowAnnotationClass ?

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
yogeshnachnani: nvm, got it

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
```diff
@@ -0,0 +1,212 @@
...
+        return methods;
+    }
+
+    /**
+     * Given a deployment unit deployment unit name retrieves properties from config file and returns them.
+     * @param deploymentUnitName
+     * @return
+     * @throws IOException
+     */
+    public YamlConfiguration getProperties(String deploymentUnitName) throws IOException {
```
yogeshnachnani: where is this used?

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
```diff
@@ -0,0 +1,212 @@
...
+     * @throws IOException
+     */
+    public YamlConfiguration getProperties(URLClassLoader urlClassLoader) throws IOException {
+        return new YamlConfiguration(urlClassLoader.getResource(CONFIG_FILE));
+    }
+
+    /**
+     * Provides {@link java.net.URLClassLoader for a given deployment unit}
+     */
+    static class ClassLoaderProvider {
```
yogeshnachnani: why inner class?

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
```diff
@@ -0,0 +1,212 @@
...
+         * @param deploymentUnitFQN
+         * @return URLClassLoader
+         * @throws java.net.MalformedURLException
+         */
+        public static URLClassLoader getClassLoader(String deploymentUnitFQN) throws MalformedURLException, FileNotFoundException {
+
+            File[] mainJars = new File(deploymentUnitFQN+"main").listFiles();
+            File[] libJars = new File(deploymentUnitFQN+"lib").listFiles();
+
+            if(mainJars == null)
```
yogeshnachnani: always keep braces { }

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
```diff
@@ -0,0 +1,212 @@
...
+                throw new FileNotFoundException("Unable to build class loader. Required directory "+deploymentUnitFQN+"main is empty.");
+
+            File configFile = new File(deploymentUnitFQN+CONFIG_FILE);
+            if(!configFile.isFile())
```
yogeshnachnani: 1. keep braces. 
2. Better to check for exists() ?

runtime/src/main/java/com/flipkart/flux/deploymentunit/ExecutableRegistryPopulator.java
```diff
@@ -0,0 +1,146 @@
...
+
+            Class classLoaderInjectorClass;
+            Class classLoaderClass = Class.forName("java.lang.ClassLoader");
+            java.lang.reflect.Method method = classLoaderClass.getDeclaredMethod("defineClass", String.class, byte[].class, int.class, int.class);
+
+            // protected method invocation
+            method.setAccessible(true);
+            try {
+                Object[] args = new Object[]{ClassLoaderInjector.class.getCanonicalName(), classBytes, 0, classBytes.length};
+                classLoaderInjectorClass = (Class) method.invoke(deploymentUnitClassLoader, args);
```
yogeshnachnani: why not do deploymentUnitClassLoader.defineClass(<name>, classBytes, 0, classBytes.length) directly?
As in, why are we extracting the "defineClass" method from the classLoader class and then invoking it?
Also, if we do need to extract the "defineClass" method, why not do it from the deploymentUnitClassLoader.getClass().getDeclaredMethod() ?

runtime/src/main/java/com/flipkart/flux/deploymentunit/ExecutableRegistryPopulator.java
```diff
@@ -0,0 +1,146 @@
...
+                    }
+
+                    Object singletonMethodOwner = getClassInstanceMethod.invoke(injectorClassInstance, method.getDeclaringClass());
+                    executableRegistry.registerTask(taskIdentifier, new ExecutableImpl(singletonMethodOwner, method, timeout, classLoader));
+
+                    //add the router name to deployment unit routers
+                    routerNames.add(method.getDeclaringClass().getName() + "_" + method.getName()); //convention: task router name is classFQN_taskMethodName
+                }
+            } catch (Exception e) {
+                LOGGER.error("Unable to populate Executable Registry for deployment unit: "+deploymentUnitName+". Exception: "+e.getMessage());
```
yogeshnachnani: For all logger usages, don't use string concatenation yourself. 
Use markup. Above would be Logger.error("Unable to.. unit: {}. Exception : ", deploymentUnitName,e);

task/src/main/java/com/flipkart/flux/impl/task/LocalJvmTask.java
```diff
@@ -72,19 +69,28 @@ public int getExecutionTimeout() {
             While this works for methods with all unique param types, it
             will fail for methods where we have mutliple params of the same type.
              */
+            ClassLoader classLoader = toInvoke.getDeploymentUnitClassLoader() != null ? toInvoke.getDeploymentUnitClassLoader() : this.getClass().getClassLoader();
```
yogeshnachnani: This should never be null, right ?

yogeshnachnani: In production I mean

runtime/src/main/java/com/flipkart/flux/deploymentunit/ExecutableRegistryPopulator.java
```diff
@@ -0,0 +1,146 @@
...
+                    injectorClassInstance = InjectorClass.getConstructor(guiceModuleClass).newInstance(classLoader.loadClass(DUModuleClassFQN).newInstance());
+                }
+
+                Set<Method> taskMethods = deploymentUnitUtil.getTaskMethods(classLoader);
+                //for every task method found in the deployment unit create an executable and keep it in executable registry
+                for(Method method : taskMethods) {
+                    String taskIdentifier = new MethodId(method).toString();
+                    Annotation taskAnnotation = method.getAnnotationsByType(TaskClass)[0];
+                    Class<? extends Annotation> annotationType = taskAnnotation.annotationType();
+                    long timeout = 1000l; //default timeout
```
yogeshnachnani: inject this using configuration. If not, at least use a constant with a todo marker

runtime/src/main/java/com/flipkart/flux/guice/module/ConfigModule.java
```diff
@@ -53,23 +65,42 @@ protected void configure() {
...
+        Set<String> routerNames = new ConcurrentHashSet<>();
+        Iterator<String> propertyKeys = routerConfig.getKeys();
+        while(propertyKeys.hasNext()) {
+            String propertyKey = propertyKeys.next();
+            String routerName = propertyKey.substring(0, propertyKey.lastIndexOf('.'));
+            routerNames.add(routerName);
+        }
+
+        //add all deployment units' routers
+        routerNames.addAll(executableRegistryPopulator.getDeploymentUnitRouterNames());
```
yogeshnachnani: does polyguice guarantee that the initialise method on ExecutableRegistryPopulator will be called before it is injected as an argument in this method?

runtime/src/main/java/com/flipkart/flux/guice/module/ConfigModule.java
```diff
@@ -53,23 +65,42 @@ protected void configure() {
...
-            Iterator<String> propertyKeys = yamlConfiguration.getKeys();
-            while (propertyKeys.hasNext()) {
-                String propertyKey = propertyKeys.next();
-                Object propertyValue = yamlConfiguration.getProperty(propertyKey);
-                LinkedBindingBuilder annotatedWith = bind(propertyValue.getClass()).annotatedWith(Names.named(propertyKey));
-                annotatedWith.toInstance(propertyValue);
-            }
-        } catch (IOException e) {
-            throw new RuntimeException(e);
+    private void bindConfigProperties() {
```
yogeshnachnani: any reason for this change?

runtime/src/main/java/com/flipkart/flux/guice/module/ConfigModule.java
yogeshnachnani: also, why do we depend on executableRegistryPopulator to get deployment unit names?
If you really need deployment units at two places, why not read independently (once) and "Provide" it wherever needed using DI ?

task/src/main/java/com/flipkart/flux/deploymentunit/ClassLoaderInjector.java
```diff
@@ -0,0 +1,61 @@
...
+        injector = Guice.createInjector();
+    }
+
+    /** Creates guice injector with specified module */
+    public ClassLoaderInjector(Module module){
+        injector = Guice.createInjector(module);
+    }
+
+    /** Given a class, returns its instance */
+    public Object getClassInstance(Class clazz) {
```
yogeshnachnani: s/getClassInstance/getInstance

task/src/main/java/com/flipkart/flux/impl/task/LocalJvmTask.java
```diff
@@ -72,19 +69,28 @@ public int getExecutionTimeout() {
...
+            Class objectMapper = classLoader.loadClass("com.fasterxml.jackson.databind.ObjectMapper");
+
             for (int i = 0 ; i < parameterTypes.length ; i++) {
                 for (EventData anEvent : events) {
-                    if(Class.forName(anEvent.getType()).equals(parameterTypes[i])) {
-                        parameters[i] = objectMapper.readValue(anEvent.getData(), Class.forName(anEvent.getType()));
+                    if(Class.forName(anEvent.getType(), true, classLoader).equals(parameterTypes[i])) {
+                        parameters[i] = objectMapper.getMethod("readValue", String.class, Class.class).invoke(objectMapper.newInstance(), anEvent.getData(), Class.forName(anEvent.getType(), true, classLoader));
```
yogeshnachnani: we are creating a new instance of objectMapper for every conversion!!! 🔫

yogeshnachnani: we need to keep a single instance of object mapper available per class loader.
I have a solution in mind, but i want you to come up with one and we can compare :)

examples/run_example.sh
shyamakirala: not sure, let me see if it's possible

runtime/pom.xml
shyamakirala: Removed it, I'm getting it as part of task module already.

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
shyamakirala: yeah, will do

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
shyamakirala: done

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
shyamakirala: This method is not called anywhere currently.

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
shyamakirala: makes sense. changed it.

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
shyamakirala: done.

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
shyamakirala: Currently it is not used anywhere.

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
shyamakirala: This class will be used by only DeploymentUnitUnit and DeploymentUnitUtil heavily depends on it. Also this class provides class loader to specific structure only. So IMO, it makes more sense to keep it as part of DeploymentUnitUtil itself.

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
shyamakirala: done

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
shyamakirala: 1. done.
2. Both serves the same purpose, if the file doesn't exist this if would be executed. I think it is good to use isFile() and throw error before itself instead of throwing at file parsing time.

runtime/src/main/java/com/flipkart/flux/deploymentunit/ExecutableRegistryPopulator.java
shyamakirala: done.

task/src/main/java/com/flipkart/flux/deploymentunit/ClassLoaderInjector.java
shyamakirala: done.

task/src/main/java/com/flipkart/flux/impl/task/LocalJvmTask.java
shyamakirala: Yes, this won't be null in production.

yogeshnachnani: so this code should never exist. If its the tests that are creating the problem then fix the tests.
Never alter production code just because we're not writing tests correctly :)

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
yogeshnachnani: remove it then

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
yogeshnachnani: remove please

runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnitUtil.java
yogeshnachnani: 2. Agreed

task/src/main/java/com/flipkart/flux/impl/task/TaskExecutor.java
```diff
@@ -78,7 +78,12 @@ protected Event run() throws Exception {
 		}
 		final Object returnObject = result.getKey();
 		if (returnObject != null) {
-			return new Event(outputeEventName, returnObject.getClass().getCanonicalName(), Event.EventStatus.triggered, stateMachineId, objectMapper.writeValueAsString(returnObject), MANAGED_RUNTIME);
+            Event event = (Event) returnObject;
+            event.setName(outputeEventName);
```
yogeshnachnani: I see that you're returning Pair<Event,..> in the LocalJvmTask & are using setters here.
Why the change?

yogeshnachnani: I get that you have to use the class loader specific object mapper in local jvm task; but why creating half an event here and half there ?

task/src/main/java/com/flipkart/flux/impl/task/LocalJvmTask.java
shyamakirala: When I wrote this code, I was under assumption that we would be using the same task module in isolation mode as well to execute and we need to have a default router which gives actors for execution. As we have discussed afterwards, task library won't be present in isolation mode, will change this code piece.

task/src/main/java/com/flipkart/flux/impl/task/TaskExecutor.java
shyamakirala: 1. To construct the full event in LocalJVMTask I need have the parameters outputeEventName, stateMachineId in LocalJVMTask but I don't want to pass them as part of parameters.
2. To construct in Task executor I don't have Deployment Unit object Mapper.
So went with this approach, if you see any other better approach I'm fine changing it.

runtime/src/main/java/com/flipkart/flux/guice/module/ConfigModule.java
shyamakirala: We were using the url to construct YamlConfiguration, as I've put it as a class field to use in multiple methods, we don't need to pass this param.

runtime/src/main/java/com/flipkart/flux/guice/module/ConfigModule.java
shyamakirala: 1. Yes
2. I need all the methods which are annotated with @Task to 1) to generate router names and 2) to put in ExecutableRegistry. To avoid scanning multiple times I'm getting the names from ExecutableRegistryPopulator. If we go with Singleton Provides then all the task methods would be cached in some place, do we really need it ? as currently we need them only twice and even if we need in future, I don't think it would be a frequent call.

runtime/src/main/java/com/flipkart/flux/deploymentunit/ExecutableRegistryPopulator.java
shyamakirala: defineClass in URLClassLoader internally calls ClassLoader's method only and it's private. So I have to fetch it to use. It accepts params String className and sun.misc.Resource, I played with it by creating a Resource from these params, but it didn't workout. If I invoke defineClass directly from ClassLoader it uses default protection domain which has only read permissions on the resource also certs as null. It works well for our purpose.

task/src/main/java/com/flipkart/flux/impl/task/TaskExecutor.java
yogeshnachnani: create another object/abstraction :) You can return a class called SerialisedEventData , or simply a string.
Use that to create the Event object in task executor :)

runtime/src/main/java/com/flipkart/flux/guice/module/ConfigModule.java
yogeshnachnani: cool.. nice

runtime/src/main/java/com/flipkart/flux/guice/module/ConfigModule.java
yogeshnachnani: Everything in Java is reference. So even if you create references in placeA and use them in placeB - the references still remain.
Similarly, if you create a reference in placeA & use them in B & C - its the same thing.

This, here, is an artificial dependency. This is clearly the case of B & C depending on A. So keep it that way

runtime/src/main/java/com/flipkart/flux/deploymentunit/ExecutableRegistryPopulator.java
yogeshnachnani: i didn't get understand. I'm saying to invoke defineClass from "deploymentUnitClassLoader" -- the param which is passed to this method. The method is ultimately invoked against that parameter.

To circumvent the permissions issue, we can extend the classLoader (or UrlClassLoader) with the relevant methods being public. Check http://download.forge.objectweb.org/asm/asm4-guide.pdf (page 17)

task/src/main/java/com/flipkart/flux/impl/task/TaskExecutor.java
shyamakirala: I thought about creating another class, but it would be a overkill as we can get away the approach I used :). If we send only a string from JVM task, we can't get the type of the class in executor.

yogeshnachnani: create new class and use :) no half-half construction please

runtime/src/main/java/com/flipkart/flux/guice/module/ConfigModule.java
shyamakirala: Yeah, everything is a reference but thought we may need to take separate copies for doing any operations.  As we won't be doing any modifications to the objects, we can just go as you suggested. changing it.

yogeshnachnani:
I am yet to look at the last few newly introduced classes (DeploymentUnitJunitRunner etc)
But overall looks fine to me - please go ahead and merge it.
It currently has merge conflicts so i can't do it over github

PR #16, opened 2016-06-21
List of events support

yogeshnachnani:
Functional - Adds support for vararg workflow methods (this enables the 2nd of our examples to work)
UI changes - support for multiple states with same name ( unblocks us from writing a decent fork-join eg)


yogeshnachnani:
will update the flux wiki with the 2nd example after one of you merges this

yogeshnachnani:
@shyamakirala - fixed the bug you pointed out yesterday

PR #21, opened 2016-09-01
JIRA: FLX-73 : Updated configmodule to read configuration from jvm param

sibasethy:
Please review the code to read flux config location from jvm param else will be default location.


common/src/main/java/com/flipkart/flux/guice/module/ConfigModule.java
```diff
@@ -37,7 +39,13 @@
 
     public ConfigModule() {
         try {
-            URL configUrl = this.getClass().getClassLoader().getResource(CONFIGURATION_YML);
+            URL configUrl = null;
+            String fluxConfigFile = System.getProperty("fluxConfigFile");
```
shyamakirala: change it to flux.configurationFile

PR #23, opened 2016-09-06
FLX-115: API to Unsideline the tasks

chandresh-pancholi:


model/src/main/java/com/flipkart/flux/domain/Status.java
```diff
@@ -19,5 +19,5 @@
  * @author regunath.balasubramanian
  */
 public enum  Status {
-    initialized, running, completed, cancelled, errored, sidelined; //possible statuses of execution
+    initialized, running, completed, cancelled, errored, sidelined, unsidelined; //possible statuses of execution
```
shyamakirala: Add unsidelined in other Status enum as well.

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -149,6 +151,25 @@ public void updateExecutionStatus(Long stateMachineId,Long taskId, Status status
     }
 
     /**
+     * Unsideline state and trigger execution states.
```
shyamakirala: Unsidelines a state and triggers its execution. :)

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -149,6 +151,25 @@ public void updateExecutionStatus(Long stateMachineId,Long taskId, Status status
...
+     * @param stateMachineId
+     * @param stateId
+     */
```
shyamakirala: add @Transactional

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -253,6 +252,21 @@ public Response getFsmGraphData(@PathParam("machineId") String machineId) throws
                 .build();
     }
 
+    /**
+     * This api unsideline states and trigger the execution.
```
shyamakirala: unsidelines a single state

PR #25, opened 2016-09-13
JIRA: FLX-34: Updated the migration runner to read liquibase resource

sibasethy:
fixed the migration script to read resources.


shyamakirala:
lgtm

PR #30, opened 2016-10-10
Using a Guice provided configuration for building the flux runtime connector

aakn:
This configuration will internally come from one of client application's Guice modules. Example of a client provider

```java
@Produces
@Singleton
FluxClientConfiguration provides FluxClientConfiguration(AppConfiguration configuration) {
    return configuration.get FluxClientConfiguration();
}
```

### DIFF
--- a/client/src/main/java/com/flipkart/flux/client/config/FluxClientConfiguration.java
+++ b/client/src/main/java/com/flipkart/flux/client/config/FluxClientConfiguration.java
@@ -2,7 +2,7 @@ package com.flipkart.flux.client.config;
 
 public class FluxClientConfiguration {
 
-  private String url;
+  private String url = "http://localhost:9998";
   private long socketTimeout = 1000;
   private long connectionTimeout = 1000;
 
--- a/client/src/main/java/com/flipkart/flux/client/config/FluxClientConfiguration.java
+++ b/client/src/main/java/com/flipkart/flux/client/config/FluxClientConfiguration.java
@@ -1,0 +1,32 @@
+package com.flipkart.flux.client.config;
+
+public class FluxClientConfiguration {
+
+  private String url;
+  private long socketTimeout = 1000;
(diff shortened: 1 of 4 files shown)


aakn:
Will update the examples with this change.

client/src/main/java/com/flipkart/flux/client/config/FluxClientConfiguration.java
```diff
@@ -0,0 +1,32 @@
+package com.flipkart.flux.client.config;
+
+public class FluxClientConfiguration {
+
+  private String url;
```
shyamakirala: pls keep "http://localhost:9998" as default value, so that to tryout in local client doesn't need to provide a config class.

aakn: done

aakn:
Tested it locally, with SFS.

client/src/main/java/com/flipkart/flux/client/intercept/WorkflowInterceptor.java
```diff
@@ -45,15 +46,15 @@
...
+    private Provider<FluxRuntimeConnector> connector;
 
     private final ObjectMapper objectMapper;
 
     public WorkflowInterceptor() {
         this.objectMapper = new ObjectMapper();
     }
 
-    public WorkflowInterceptor(LocalContext localContext, FluxRuntimeConnector connector) {
+    public WorkflowInterceptor(LocalContext localContext, Provider<FluxRuntimeConnector> connector) {
```
shyamakirala: connectorProvider instead of connector ?

aakn: sure

aakn: done

PR #33, opened 2016-10-17
Configurable Retry support

shyamakirala:
1. Added FluxRetriableException, on its occurrence Flux Runtime tries to do task execution retry if the task has not exhausted the configured number of retries

### DIFF
--- a/client/src/main/java/com/flipkart/flux/client/exception/FluxRetriableException.java
+++ b/client/src/main/java/com/flipkart/flux/client/exception/FluxRetriableException.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2012-2016, the original author or authors.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.flipkart.flux.client.exception;
+
+/**
+ * <code>FluxRetriableException</code> indicates an Exception on occurrence of it Flux Runtime retries task execution if
+ * the task has not exhausted the configured number of retries.
+ *
+ * For ex:
+ *      {@literal @}Task(version = 1, retires = 2, timeout = 1000)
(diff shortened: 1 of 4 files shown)


task/src/main/java/com/flipkart/flux/impl/task/AkkaTask.java
```diff
@@ -131,10 +132,30 @@ public void onReceive(Object message) throws Exception {
...
                     } else {
-                        // mark the task outcome as execution failure
-                        fluxRuntimeConnector.updateExecutionStatus(
-                                new ExecutionUpdateData(taskAndEvent.getStateMachineId(), taskAndEvent.getTaskId(), Status.errored,
-                                        taskAndEvent.getRetryCount(), taskAndEvent.getCurrentRetryCount(), hre.getMessage()));
+                        //check if the exception hierarchy has FluxRetriableException, if yes trigger retry
+                        boolean isFluxRetriableException = false;
+                        Throwable cause = hre.getCause();
+                        while(cause != null) {
+                            if(cause.getClass().getName().equals(FluxRetriableException.class.getName())) {
```
gauravAshok: It will not work, if the task throws an exception of type that extends FluxRetriableException.

client/src/main/java/com/flipkart/flux/client/exception/FluxRetriableException.java
```diff
@@ -0,0 +1,42 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.flipkart.flux.client.exception;
+
+/**
+ * <code>FluxRetriableException</code> indicates an Exception on occurrence of it Flux Runtime retries task execution if
```
gauravAshok: Fullstop missing after "indicates an exception".

client/src/main/java/com/flipkart/flux/client/exception/FluxRetriableException.java
```diff
@@ -0,0 +1,42 @@
...
+ * the task has not exhausted the configured number of retries.
+ *
+ * For ex:
+ *      {@literal @}Task(version = 1, retires = 2, timeout = 1000)
```
gauravAshok: spelling mistake in "retries"

task/src/main/java/com/flipkart/flux/impl/task/AkkaTask.java
gauravAshok: There is a todo in the comment above about the RetriableException which can be removed.

PR #35, opened 2016-10-18
Injecting the object mapper in the flux client

aakn:
updated UT's (and added constructors for the same)

### DIFF
--- a/client/src/main/java/com/flipkart/flux/client/intercept/WorkflowInterceptor.java
+++ b/client/src/main/java/com/flipkart/flux/client/intercept/WorkflowInterceptor.java
@@ -14,6 +14,8 @@
 
 package com.flipkart.flux.client.intercept;
 
+import com.google.common.annotations.VisibleForTesting;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.flipkart.flux.api.EventData;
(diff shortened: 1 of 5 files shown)


aakn:
Tested this locally with SFS

client/src/main/java/com/flipkart/flux/client/runtime/FluxRuntimeConnectorHttpImpl.java
```diff
@@ -50,9 +53,14 @@
     private final String fluxEndpoint;
     private final ObjectMapper objectMapper;
 
+    @VisibleForTesting
```
shyamakirala: Same as above. Is it needed ?

client/src/main/java/com/flipkart/flux/client/intercept/WorkflowInterceptor.java
```diff
@@ -50,16 +52,18 @@
     @Inject
     private Provider<FluxRuntimeConnector> connectorProvider;
 
-    private final ObjectMapper objectMapper;
+    @Inject
+    private Provider<ObjectMapper> objectMapper;
```
shyamakirala: objectMapperProvider ? :)

client/src/main/java/com/flipkart/flux/client/intercept/WorkflowInterceptor.java
```diff
@@ -50,16 +52,18 @@
...
 
     public WorkflowInterceptor() {
-        this.objectMapper = new ObjectMapper();
     }
 
-    public WorkflowInterceptor(LocalContext localContext, Provider<FluxRuntimeConnector> connectorProvider) {
+    @VisibleForTesting
```
aakn: this is used to specify that a public method is only publicly available for testing. Should not be used in prod flows.

client/src/main/java/com/flipkart/flux/client/intercept/WorkflowInterceptor.java
aakn: done

PR #37, opened 2016-11-04
FLX-146: adding hystrix thread pool size config for a task execution …

gauravAshok:
…in flux_config.yml

### DIFF
--- a/runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnit.java
+++ b/runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnit.java
@@ -18,7 +18,7 @@ import com.flipkart.polyguice.config.YamlConfiguration;
 import org.apache.commons.io.IOUtils;
 
 import java.lang.reflect.Method;
-import java.util.Set;
+import java.util.*;
 
 /**
  * <code>DeploymentUnit</code> represents a Deployment Unit.
@@ -111,5 +111,15 @@ public class DeploymentUnit {
 
     public Object getInjectorClassInstance() {
         return injectorClassInstance;
+    }
+
+    public Map<String, Object> getTaskConfiguration(String taskPrefix) {
+        Iterator<String> allKeys = configuration.getKeys("taskConfig." + taskPrefix);
+        Map<String, Object> config = new HashMap<>();
+        while(allKeys.hasNext()) {
+            String nextKey = allKeys.next();
+            config.put(nextKey.replaceFirst("taskConfig." + taskPrefix + ".", ""), configuration.getProperty(nextKey));
+        }
+        return config;
     }
 }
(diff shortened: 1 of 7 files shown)


runtime/src/main/java/com/flipkart/flux/deploymentunit/DeploymentUnit.java
```diff
@@ -112,4 +112,14 @@ public Object getObjectMapperInstance() {
     public Object getInjectorClassInstance() {
         return injectorClassInstance;
     }
+
+    public Map<String, Object> getTaskConfiguration(String taskPrefix) {
+        Iterator<String> allKeys = configuration.getKeys("taskConfig." + taskPrefix);
```
shyamakirala: instead of getting keys and removing "taskConfig.taskPrefix", you can use configuration.subset("taskConfig.taskPrefix") which returns Configuration object with keys not having the prefix.

runtime/src/test/resources/flux_config.yml
```diff
@@ -0,0 +1,7 @@
+#guiceModuleClass: "com.flipkart.flux.examples.externalevents.SellerVerificationWorkflowModule"
+workflowClasses: ["com.flipkart.flux.dao.TestWorkflow",
+                  "com.flipkart.flux.integration.SimpleWorkflow"]
+taskConfig:
+  com.flipkart.flux.integration.SimpleWorkflow_simpleStringReturningTask:
+    execution:
```
shyamakirala: I think "execution_concurrency" would be good, as there won't be many properties under execution.

runtime/src/test/resources/flux_config.yml
```diff
@@ -0,0 +1,7 @@
+#guiceModuleClass: "com.flipkart.flux.examples.externalevents.SellerVerificationWorkflowModule"
```
shyamakirala: remove the commented line

runtime/src/main/java/com/flipkart/flux/deploymentunit/ExecutableRegistryPopulator.java
```diff
@@ -123,10 +124,16 @@ public void run() {
                         }
                     }
 
-                    String taskIdentifier = new MethodId(method).toString() + _VERSION + version;
+                    MethodId methodId = new MethodId(method);
+                    String taskIdentifier = methodId.toString() + _VERSION + version;
+
+                    /* get concurrency config for this task */
+                    Map<String, Object> taskConfig = deploymentUnit.getTaskConfiguration(methodId.getPrefix());
```
shyamakirala: Instead of having a method in DeploymentUnit and calling it, I think we can retrieve once and keep the configuration object here itself. 

org.apache.commons.configuration.Configuration taskConfiguration = deploymentUnit.getConfiguration().subset("taskconfig");

taskConfiguration.getProperty(methodId.getPrefix() + ".executionConcurrency")

PR #41, opened 2016-11-14
Leave akka cluster in shutdown hook

shyamakirala:
1. Send cluster leave message from shutdown hook
2. Changes to redriver delete logic

### DIFF
--- a/redriver/src/main/java/com/flipkart/flux/redriver/scheduler/MessageScheduler.java
+++ b/redriver/src/main/java/com/flipkart/flux/redriver/scheduler/MessageScheduler.java
@@ -25,7 +25,6 @@ import javax.inject.Named;
 import javax.inject.Singleton;
 import java.util.Date;
 import java.util.List;
-import java.util.Optional;
 import java.util.PriorityQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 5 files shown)


redriver/src/main/java/com/flipkart/flux/redriver/scheduler/MessageScheduler.java
```diff
@@ -68,17 +67,12 @@ public void addMessage(ScheduledMessage scheduledMessage) {
...
-                such cases by consulting its own DB.
-             */
-            this.messages.remove(message);
-            this.messageManagerService.scheduleForRemoval(message);
-        }
+        /* It is important that we delete from priority queue first. Its okay even if we the schedule for removal call fails.
+            At most we will send a message that was not supposed to be sent but the receiver should be able to handle
+            such cases by consulting its own DB.
+        */
+        this.messages.removeIf((m) -> m.getTaskId().equals(taskId));
```
gauravAshok: Hey, can we use linkedHashSet instead of PriorityQueue. If tasks are registered roughly in the order of their executionTime then the task with the earliest execTime should be at the front of linkedList.

gauravAshok:
Looks good. We will look into the priority queue, if it creates problem in perf.

PR #42, opened 2016-11-14
Extracted the http client into a separate interface

aakn:
**Changelog**
* Decoupled the http client building from the FluxRuntimeConnector, extracted the http client POST logic to a separate class
* The apache http client is now build via guice
* Adding a few log statements to find the exact error


client/src/main/java/com/flipkart/flux/client/runtime/FluxHttpClient.java
```diff
@@ -0,0 +1,6 @@
+package com.flipkart.flux.client.runtime;
+
+public interface FluxHttpClient {
+
+    void postOverHttp(Object dataToPost, String pathSuffix);
```
shyamakirala: while returning void works for our current use cases, will face issues when we get the response after doing post. This same API is used by both Client and Task runtime to connect to Flux runtime.

client/src/main/java/com/flipkart/flux/client/runtime/FluxRuntimeConnectorHttpImpl.java
```diff
@@ -19,175 +19,83 @@
...
-            final ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
-            objectMapper.writeValue(byteArrayOutputStream, dataToPost);
-            httpPostRequest.setEntity(new ByteArrayEntity(byteArrayOutputStream.toByteArray(), ContentType.APPLICATION_JSON));
-            httpResponse = closeableHttpClient.execute(httpPostRequest);
-            final int statusCode = httpResponse.getStatusLine().getStatusCode();
-            if (statusCode >= Response.Status.OK.getStatusCode() && statusCode < Response.Status.MOVED_PERMANENTLY.getStatusCode() ) {
-                // all is well, TODO write a trace level log
-            } else {
-                // TODO: log status line here
-                throw new RuntimeCommunicationException("Did not receive a valid response from Flux core");
```
shyamakirala: Closing the connection here should solve the leak issue with having a thread cleanup expired and idle connections periodically.

client/src/main/java/com/flipkart/flux/client/runtime/FluxHttpClient.java
aakn: What we can do is, add another method which takes the type to be returned as parameter (or can be done via generics as well), and use object mapper to convert the response.

client/src/main/java/com/flipkart/flux/client/runtime/FluxRuntimeConnectorHttpImpl.java
aakn: it does get closed in the finally block.

aakn: yeah, we can add the connection cleanup code, [which will run in background?] to clean up the close wait connections

client/src/main/java/com/flipkart/flux/client/runtime/FluxHttpClient.java
aakn: Will the postOverHttp be used by the task runtime? or does the flux runtime connector get used by the runtime?

PR #43, opened 2016-11-14
Adding api resource for managing rotation status, cluster nodes, fetc…

gauravAshok:
Added resource apis to
- manage rotation status
- to take out the node out of cluster.
- to fetch all errored states for a range of stateMachine Ids.


runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -34,6 +34,7 @@
 import com.google.inject.Inject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.util.StringUtils;
```
shyamakirala: unused import

PR #44, opened 2016-11-22
Reworked redriver. 

gauravAshok:
Earlier, if the node holding the singleton of MessageScheduler were to go down, the cluster would stop processing any request, as the register/deregister calls would stuck.

* Moved register/deregister methods out of the messageScheduler. Now every node serves its own register/deregister requests.
* RedriverRegisterImpl takes care of the register/deregister calls by directing it to MessageManagerService.
* Renamed messageScheduler to RedriverService. It's job now only is to periodically check DB for oldest messages and redrive it if there is a need.


redriver/src/main/java/com/flipkart/flux/redriver/service/MessageManagerService.java
```diff
@@ -39,70 +40,74 @@
...
     public MessageManagerService(MessageDao messageDao,
+                                 @Named("redriver.noOfPersistenceWorkers") int noOfPersistenceWorkers,
                                  @Named("redriver.batchdelete.intervalms") Integer batchDeleteInterval,
                                  @Named("redriver.batchdelete.batchsize") Integer batchSize) {
         this.messageDao = messageDao;
-        this.batchDeleteInterval = Long.valueOf(batchDeleteInterval);
+        this.batchDeleteInterval = batchDeleteInterval;
         this.batchSize = batchSize;
         this.messagesToDelete = new ConcurrentLinkedQueue<>();
         scheduledExecutorService =
```
shyamakirala: scheduledDeletionService ?

redriver/src/main/java/com/flipkart/flux/redriver/service/MessageManagerService.java
```diff
@@ -39,70 +40,74 @@
...
-            new InstrumentedScheduledExecutorService(Executors.newScheduledThreadPool(2), SharedMetricRegistries.getOrCreate(METRIC_REGISTRY_NAME));
+            new InstrumentedScheduledExecutorService(Executors.newScheduledThreadPool(2), SharedMetricRegistries.getOrCreate(METRIC_REGISTRY_NAME), scheduledExectorSvcName);
+        persistenceExecutorService = Executors.newFixedThreadPool(noOfPersistenceWorkers);
```
shyamakirala: I think we should change it to InstrumentedExecutorService to get metrics

redriver/src/main/java/com/flipkart/flux/redriver/service/RedriverService.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+import static com.flipkart.flux.Constants.METRIC_REGISTRY_NAME;
+
+/**
+ * The service uses a scheduler to read the oldest message with fixed delay and redrives them if necessary. The task will
+ * be scheduled for execution if the current time is greater than or equal to scheduledTime.
```
shyamakirala: @author :)

runtime/src/main/resources/packaged/configuration.yml
```diff
@@ -48,7 +48,9 @@ redriver:
   batchdelete:
     batchsize: 50
     intervalms: 1000
-  noOfRedriverWorkers : 50
+  batchread:
+    batchsize: 100
+    intervalms: 20000
```
shyamakirala: we may need to reduce it. Will do a perf before putting in prod, and fix this.

redriver/src/main/java/com/flipkart/flux/redriver/service/RedriverService.java
```diff
@@ -0,0 +1,108 @@
...
+        int offset = 0;
+        List<ScheduledMessage> messages;
+        long now = System.currentTimeMillis();
+        do {
+            messages = messageService.retrieveOldest(offset, batchSize);
+            messages.stream().filter(e -> e.getScheduledTime() < now).forEach(e -> redriverRegistry.redriveTask(e.getTaskId()));
+            offset += batchSize;
+            // get next batch if we found batchSize tasks and all were redriven
+        } while (messages.size() == batchSize && messages.get(messages.size() - 1).getScheduledTime() < now);
+    }
```
gauravAshok: you mean, if there are only few total msgs in db, then we clear them up regardless of their timings. OR if we found few msgs that can be redriven then we should do that. the latter is already happening.

redriver/src/main/java/com/flipkart/flux/redriver/service/MessageManagerService.java
gauravAshok: are you suggesting the class name?

redriver/src/main/java/com/flipkart/flux/redriver/service/RedriverService.java
shyamakirala: was talking about the second case. My bad, didn't see it.

shyamakirala:
Looks good.

PR #45, opened 2016-11-25
Adding support for routing db calls to different dataSource based on an annotation

gauravAshok:
* Added a SessionFactoryContext which holds the current factory to be used in a  transaction.
* TransactionInterceptor uses the annotation on the invoked DAO method to appropriately set a SessionFactory in the current context.
* Enables to route read-only/batch read apis to slave db.
* Increased range for searching errored/sidelined states to 1Million from 1Thousand.


shyamakirala:
This branch has conflicts with Master, take a pull.

PR #48, opened 2016-12-05
Changed Singleton router to local routers.

shyamakirala:
Changes include:
1. Spawning local router with its own actors on each box, instead of having a cluster singleton router.
2. Local retries handled by Akka scheduler with exponential backoff time
3. Spawing the same no.of Actors and Hystrix threads for a task. If the executionConcurrency is not mentioned in deployment unit config file, it takes default value from runtime's configuration.yml routers.default.instancesPerNode
4. Redriver delay for a task has been changed to support exponential backoff
5. After the execution of a task, the event should be posted to Flux Runtime's ELB, not the localhost. To do this taking flux runtime url from System properties. When starting flux it has to be passed as -Dflux.runtimeUrl=http://url:port
Fixed:
1. While doing retries using Redriver, if a task has 5 retries and 3 retries are already happened on other box, we should try only remaining 2 executions, not all 5
2. Cleaning up the tasks from the redriver DB, which can't be redrived


runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -240,6 +243,9 @@ public void redriveTask(Long taskId) {
         if(isTaskRedrivable(state.getStatus()) && state.getAttemptedNoOfRetries() < state.getRetryCount()) {
             logger.info("Redriving a task with Id: {} for state machine: {}", state.getId(), state.getStateMachineId());
             executeStates(state.getStateMachineId(), Collections.singleton(state));
+        } else {
+            //cleanup the tasks which can't be redrived from redriver db
+            this.redriverRegistry.deRegisterTask(taskId);
         }
     }
```
gauravAshok: We should deregister even in the case when updateExecutionStates is called with  fatal "errored" state since we don't want that to be redriven.

runtime/src/main/resources/packaged/webapps/dashboard/WEB-INF/pages/fsmview.ftl
```diff
@@ -45,6 +45,10 @@
                         <th align="left" style="border-top: none; vertical-align: middle;">Correlation Id:</th>
                         <td id="correlationId" align="left" style="border-top: none; vertical-align: middle;"></td>
                     </tr>
```
gauravAshok: We should also expose the workflow name, so that clients can use API which rely on the name.

task/src/main/java/com/flipkart/flux/impl/task/AkkaTask.java
```diff
@@ -77,121 +76,127 @@
...
+                        FailureType ft = hre.getFailureType();
+                        // we signal a timeout for any of Timeout, ThreadPool Rejection or Short-Circuit - all of these may go through with time and retry
+                        if (ft.equals(FailureType.REJECTED_THREAD_EXECUTION) || ft.equals(FailureType.SHORTCIRCUIT) || ft.equals(FailureType.TIMEOUT)) {
+                            // update flux runtime with task outcome as timeout
                             fluxRuntimeConnector.updateExecutionStatus(new ExecutionUpdateData(
                                     taskAndEvent.getStateMachineId(), taskAndEvent.getTaskId(), Status.errored, taskAndEvent.getRetryCount(),
-                                    taskAndEvent.getCurrentRetryCount(), cause.getMessage()));
-                            throw new FluxError(FluxError.ErrorType.retriable, cause.getMessage(), null, false,
+                                    taskAndEvent.getCurrentRetryCount(), "Execution timeout for : " + task.getName()));
+                            throw new FluxError(FluxError.ErrorType.timeout, "Execution timeout for : " + task.getName(),
```
gauravAshok: We should expose the failure_type in the message.

task/src/main/java/com/flipkart/flux/impl/task/AkkaTask.java
```diff
@@ -77,121 +76,127 @@
...
                         } else {
-                            // mark the task outcome as execution failure
-                            fluxRuntimeConnector.updateExecutionStatus(
-                                    new ExecutionUpdateData(taskAndEvent.getStateMachineId(), taskAndEvent.getTaskId(), Status.errored,
-                                            taskAndEvent.getRetryCount(), taskAndEvent.getCurrentRetryCount(), hre.getMessage()));
+                            //check if the exception hierarchy has FluxRetriableException, if yes trigger retry
+                            boolean isFluxRetriableException = false;
+                            Throwable cause = hre.getCause();
+                            Throwable rootCause = cause;
+                            while (cause != null) {
```
gauravAshok: Tracking the cause was achievable using just 1 variable.

task/src/main/java/com/flipkart/flux/impl/task/AkkaTask.java
```diff
@@ -77,121 +76,127 @@
...
+                                if (cause.getClass().getName().equals(FluxRetriableException.class.getName())) {
+                                    isFluxRetriableException = true;
+                                    break;
+                                }
+                                rootCause = cause;
+                                cause = cause.getCause();
+                            }
+                            if (isFluxRetriableException) {
+                                // mark the task outcome as execution failure
+                                fluxRuntimeConnector.updateExecutionStatus(new ExecutionUpdateData(
```
gauravAshok: once we have 1 variable to track cause, updateExecState on line 141, 149 can be moved out of if-else block.

task/src/main/java/com/flipkart/flux/impl/task/AkkaTask.java
```diff
@@ -77,121 +76,127 @@
...
+                                message,
+                                getContext().system().dispatcher(), null);
+
+                    } else {
+                        logger.warning("Aborting retries for Task Id : {}. Retry count exceeded : {}", fe.getExecutionContextMeta().getTaskId(),
+                                fe.getExecutionContextMeta().getAttemptedNoOfRetries());
+                        // update the Flux runtime to mark the Task as sidelined
+                        fluxRuntimeConnector.updateExecutionStatus(new ExecutionUpdateData(fe.getExecutionContextMeta().getStateMachineId(),
+                                fe.getExecutionContextMeta().getTaskId(), Status.sidelined, fe.getExecutionContextMeta().getMaxRetries(),
+                                fe.getExecutionContextMeta().getAttemptedNoOfRetries(), fe.getMessage()));
```
gauravAshok: This call should remove the message from redriver.

task/src/main/java/com/flipkart/flux/impl/task/AkkaTask.java
gauravAshok: This method is huge. We should move the call to updateExecutionStatus into another helper method. same also can be done to the throwing FluxError.

gauravAshok:
Looks good. Will merge.

PR #56, opened 2017-02-14
Adding a check to not execute a task if it is already completed

shyamakirala:


shyamakirala:
Ignore the Travis ci build failures as it's not yet setup properly.

PR #58, opened 2017-02-28
Add Task Status Level Metrics

amitkumar94:


runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -84,7 +84,7 @@ public WorkFlowExecutionController(EventsDAO eventsDAO, StateMachinesDAO stateMa
 
         final List<String> triggeredEvents = eventsDAO.findTriggeredEventsNamesBySMId(stateMachine.getId());
         Set<State> initialStates = context.getInitialStates(new HashSet<>(triggeredEvents));
-        executeStates(stateMachine.getId(), initialStates);
+        executeStates(stateMachine.getId(), stateMachine.getName(), initialStates);
```
shyamakirala: Instead of sending name as one more parameter, directly pass state machine object itself, and take the required params there.

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -154,14 +154,17 @@ public void updateExecutionStatus(Long stateMachineId,Long taskId, Status status
      */
     public void unsidelineState(Long stateMachineId, Long stateId) {
         State state = this.statesDAO.findById(stateId);
+        StateMachine stateMachine = retrieveStateMachine(stateMachineId);
```
shyamakirala: Here we can make a single call to DB to fetch state machine and after that we can loop through Set<State> of state machine to obtain state object.

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -181,18 +184,18 @@ private StateMachine retrieveStateMachineByCorrelationId(String correlationId) {
     }
 
     /**
-     * Wrapper function on {@link #executeStates(Long, Set, Event)} which triggers the execution of executableStates using Akka router.
+     * Wrapper function on {@link #executeStates(Long, String, Set, Event)} which triggers the execution of executableStates using Akka router.
      */
-    private void executeStates(Long stateMachineInstanceId, Set<State> executableStates) {
-        executeStates(stateMachineInstanceId, executableStates, null);
+    private void executeStates(Long stateMachineInstanceId, String stateMachineName, Set<State> executableStates) {
```
shyamakirala: Take state machine object as mentioned earlier

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -260,8 +263,9 @@ private StateMachine retrieveStateMachine(Long stateMachineInstanceId) {
     public void redriveTask(Long taskId) {
         State state = statesDAO.findById(taskId);
         if(state != null && isTaskRedrivable(state.getStatus()) && state.getAttemptedNoOfRetries() < state.getRetryCount()) {
+            StateMachine stateMachine = retrieveStateMachine(state.getStateMachineId());
```
shyamakirala: Here also we can get away with single DB call.

task/src/main/java/com/flipkart/flux/impl/message/TaskAndEvents.java
```diff
@@ -114,6 +120,7 @@ public boolean equals(Object o) {
         // Probably incorrect - comparing Object[] arrays with Arrays.equals
         if (!Arrays.equals(events, that.events)) return false;
         if (!stateMachineId.equals(that.stateMachineId)) return false;
+        // Didn't include stateMachineName in the equals, as two taskEvent can have the same stateMachineName
```
shyamakirala: agreed. we don't need stateMachineName. Remove the comment as well, not needed.

task/src/main/java/com/flipkart/flux/impl/message/TaskAndEvents.java
```diff
@@ -124,6 +131,7 @@ public int hashCode() {
         result = 31 * result + (events != null ? Arrays.hashCode(events) : 0);
         result = 31 * result + taskId.hashCode();
         result = 31 * result + stateMachineId.hashCode();
+        result = 31 * result + stateMachineName.hashCode();
```
shyamakirala: Always keep hashCode and Equals with same params. So remove this.

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
amitkumar94: we don't have state machine id, unlike earlier case, we derive that from the state

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -153,15 +153,27 @@ public void updateExecutionStatus(Long stateMachineId,Long taskId, Status status
      * @param stateId
      */
     public void unsidelineState(Long stateMachineId, Long stateId) {
-        State state = this.statesDAO.findById(stateId);
+        State askedState = null;
+        StateMachine stateMachine = retrieveStateMachine(stateMachineId);
+        for (State state: stateMachine.getStates()){
```
shyamakirala: this will throw null pointer if stateMachine not found. Keep the null check before.

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -181,30 +193,30 @@ private StateMachine retrieveStateMachineByCorrelationId(String correlationId) {
     }
 
     /**
-     * Wrapper function on {@link #executeStates(Long, Set, Event)} which triggers the execution of executableStates using Akka router.
+     * Wrapper function on {@link #executeStates(StateMachine, Set)} which triggers the execution of executableStates using Akka router.
```
shyamakirala: wrapper on executeStates(StateMachine, Set, Event)

PR #59, opened 2017-03-14
Task Queue Metrics

hoodakaushal:
Added counters for number of tasks in queue for every task type.


runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -237,6 +242,7 @@ private void executeStates(StateMachine stateMachine, Set<State> executableState
                 ActorRef router = this.routerRegistry.getRouter(routerName);
 
                 router.tell(msg, ActorRef.noSender());
+                metricsClient.incCounter(msg.getTaskName());
```
shyamakirala: keep the counter name in the below format

statemachine.<sm_name>.task.<task_name>.queuesize

task/src/main/java/com/flipkart/flux/impl/task/AkkaTask.java
```diff
@@ -104,6 +108,12 @@ public void onReceive(Object message) throws Exception {
                 if (task != null) {
                     try {
                         // update the Flux runtime with status of the Task as running
+                        metricsClient.decCounter(new StringBuilder().
```
shyamakirala: Please move this to line 100 (after TaskAndEvents taskAndEvent = (TaskAndEvents) message;). Just after receiving the message the counter should be decremented.

PR #60, opened 2017-03-23
API to give initialized tasks of a workflow

hoodakaushal:
Added API to retrieve initialized tasks of a workflow in a given time range.

### DIFF
--- a/runtime/src/main/java/com/flipkart/flux/dao/iface/StatesDAO.java
+++ b/runtime/src/main/java/com/flipkart/flux/dao/iface/StatesDAO.java
@@ -15,10 +15,7 @@ package com.flipkart.flux.dao.iface;
 
 import com.flipkart.flux.domain.State;
 import com.flipkart.flux.domain.Status;
-import com.flipkart.flux.persistence.DataSourceType;
-import com.flipkart.flux.persistence.SelectDataSource;
 
-import javax.transaction.Transactional;
 import java.sql.Timestamp;
 import java.util.List;
 
--- a/runtime/src/main/java/com/flipkart/flux/dao/iface/StatesDAO.java
+++ b/runtime/src/main/java/com/flipkart/flux/dao/iface/StatesDAO.java
@@ -15,7 +15,10 @@ package com.flipkart.flux.dao.iface;
 
 import com.flipkart.flux.domain.State;
 import com.flipkart.flux.domain.Status;
+import com.flipkart.flux.persistence.DataSourceType;
+import com.flipkart.flux.persistence.SelectDataSource;
 
+import javax.transaction.Transactional;
 import java.sql.Timestamp;
 import java.util.List;
 
@@ -51,4 +54,9 @@ public interface StatesDAO {
      * the given range fromTime and toTime with optional taskName parameter
      */
     List findErroredStates(String stateMachineName, Timestamp fromTime, Timestamp toTime, String taskName);
+
+    /**
+     * Returns all states with initialized status for a particular state machine name created in given time range.
+     */
+    List findInitializedStates(String stateMachineName, Timestamp fromTime, Timestamp toTime);
 }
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/com/flipkart/flux/dao/iface/StatesDAO.java
```diff
@@ -15,7 +15,10 @@
 
 import com.flipkart.flux.domain.State;
 import com.flipkart.flux.domain.Status;
+import com.flipkart.flux.persistence.DataSourceType;
+import com.flipkart.flux.persistence.SelectDataSource;
```
shyamakirala: unused

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -365,8 +365,33 @@ public Response getErroredStatesByTime(@PathParam("stateMachineName") String sta
...
+    }
+
+    /**
+     * Retrieves all initialized states for the given range of time for a particular state machine name.
+     * @return json containing list of [state machine id, state id, status]
+     */
+    @GET
+    @Path("/{stateMachineName}/states/initializedbytime")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getInitializedStatesByTime(@PathParam("stateMachineName") String stateMachineName,
```
shyamakirala: I think it's good to abstract API as well. Keep a generic API getStatesByStatusesAndTime, and accept statuses as Query param.

hoodakaushal: Should I add an extra generic API, or replace these ones? I did consider the latter, but then it would break for people who are using the current errored API.

shyamakirala: Replace it, we can coordinate with teams for deployment.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -344,17 +344,18 @@ public Response getErroredStates(@PathParam("stateMachineName") String stateMach
...
 
     /**
-     * Retrieves all errored states for the given range of time for a particular state machine name.
+     * Retrieves all states for the given range of time for a particular state machine name.
+     * Will also filter by status if it is given.
      * @return json containing list of [state machine id, state id, status]
      */
     @GET
-    @Path("/{stateMachineName}/states/erroredbytime")
+    @Path("/{stateMachineName}/states/bytime")
```
shyamakirala: states/listbytime ?

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -344,17 +344,18 @@ public Response getErroredStates(@PathParam("stateMachineName") String stateMach
...
     @Produces(MediaType.APPLICATION_JSON)
     public Response getErroredStatesByTime(@PathParam("stateMachineName") String stateMachineName,
```
shyamakirala: change method name as well

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -365,8 +366,19 @@ public Response getErroredStatesByTime(@PathParam("stateMachineName") String sta
         if(fromTimestamp.after(toTimestamp)) {
             return Response.status(Response.Status.BAD_REQUEST).entity("fromTime: " + fromTime + " should be before toTime: " + toTime).build();
         }
-
-        return Response.status(200).entity(statesDAO.findErroredStates(stateMachineName, fromTimestamp, toTimestamp, stateName)).build();
+        List<Status> statuses = new ArrayList<Status>();
+        if ("errored".equals(status)) {
+            statuses = Arrays.asList(Status.errored, Status.cancelled, Status.sidelined);
```
shyamakirala: accept statuses as list from user itself, so that they can pass whatever they want

PR #62, opened 2017-03-30
Scheduled event support

shyamakirala:
Now flux can accept events with a "triggerTime" epoch timestamp, and process the event at that particular time.

To submit scheduled event, user need to call the below API.
new FluxRuntimeConnectorHttpImpl(connectionTimeout, socketTimeout, fluxEndpoint).
                submitScheduledEvent(eventName, data, correlationId, eventSource, triggerTime);


shyamakirala:
Don't merge it yet

runtime/src/test/java/com/flipkart/flux/resource/StateMachineResourceTest.java
```diff
@@ -208,6 +213,63 @@ public void testPostEvent_againstNonExistingCorrelationId() throws Exception {
...
+
+        //request with searchField param having value other than correlationId
+        final HttpResponse<String> eventPostResponseWithWrongTag = Unirest.post(STATE_MACHINE_RESOURCE_URL+SLASH+"magic_number_1"+"/context/events?searchField=dummy&triggerTime=123").header("Content-Type","application/json").body(eventJson).asString();
+        assertThat(eventPostResponseWithWrongTag.getStatus()).isEqualTo(Response.Status.BAD_REQUEST.getStatusCode());
+    }
+
+    @Test
+    public void testPostScheduledEvent_withCorrelationId() throws Exception {
+        //create state machine
+        String stateMachineDefinitionJson = IOUtils.toString(this.getClass().getClassLoader().getResourceAsStream("state_machine_definition.json"));
```
amitkumar94: StateMachineCreation for each test case can be put in one place

scheduler/src/main/java/com/flipkart/flux/eventscheduler/model/ScheduledEvent.java
```diff
@@ -0,0 +1,135 @@
...
+        if (!eventName.equals(that.eventName)) return false;
+
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = correlationId.hashCode();
+        result = 31 * result + eventName.hashCode();
+        result = 31 * result + (int) (scheduledTime ^ (scheduledTime >>> 32));
```
amitkumar94: Didn't understand this logic

scheduler/src/main/java/com/flipkart/flux/eventscheduler/service/EventSchedulerService.java
```diff
@@ -0,0 +1,126 @@
...
+        do {
+            events = eventSchedulerDao.retrieveOldest(offset, batchSize);
+            events.stream().filter(e -> e.getScheduledTime() <= now).forEach(e -> {
+                try {
+                    EventData eventData = objectMapper.readValue(e.getEventData(), EventData.class);
+                    eventSchedulerRegistry.triggerEvent(e.getEventName(), eventData.getData(), e.getCorrelationId(), eventData.getEventSource());
+                    eventSchedulerRegistry.deregisterEvent(e.getCorrelationId(), e.getEventName());
+                } catch (Exception ex) {}
+            });
+            offset += batchSize;
```
amitkumar94: shouldn't this be offset += list.size()

amitkumar94: shouldn't the offset be always 0, as we are deleting the rows from the database along. So the new query will not have old rows, and we don't need to increment the offset.

scheduler/src/test/java/com/flipkart/flux/eventscheduler/service/EventSchedulerServiceTest.java
```diff
@@ -0,0 +1,115 @@
...
+                thenReturn(Arrays.asList(new ScheduledEvent("smCorId", "event_name1", now-2, "{\"name\":\"event_name1\",\"type\":\"some_type\",\"data\":\"data\",\"eventSource\":\"internal\"}"),
+                        new ScheduledEvent("smCorId", "event_name2", now + 10000, "{\"name\":\"event_name2\",\"type\":\"some_type\",\"data\":\"data\",\"eventSource\":\"internal\"}"))).
+                thenReturn(Arrays.asList(new ScheduledEvent("smCorId", "event_name2", now + 10000, "{\"name\":\"event_name2\",\"type\":\"some_type\",\"data\":\"data\",\"eventSource\":\"internal\"}")));
+
+        eventSchedulerService.start();
+
+        Thread.sleep(300);
+        verify(eventSchedulerRegistry).triggerEvent("event_name1", "data", "smCorId", "internal");
+        verify(eventSchedulerRegistry).deregisterEvent("smCorId", "event_name1");
+        Thread.sleep(500);
```
amitkumar94: shouldn't the sleep time here more than 1000ms, to let evenScheduler run few more times and still not consume the second Scheduled Event

shyamakirala: the first run, and next few runs of scheduler results in the same outcome. So thought why to keep more sleep time, which is not necessary. Anyway increasing it a bit to have 2 runs.

scheduler/src/main/java/com/flipkart/flux/eventscheduler/service/EventSchedulerService.java
shyamakirala: Thanks, fixed it.

runtime/src/test/java/com/flipkart/flux/resource/StateMachineResourceTest.java
shyamakirala: As we perform different operations in each test case (submitting different events for ex.), it is not possible to have a single state machine for all.

PR #63, opened 2017-04-12
Logging Tags

hoodakaushal:
Add some contextual info to logs via MDC for better tracking of workflows/tasks.

### DIFF
--- a/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
+++ b/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
@@ -30,6 +30,7 @@ import com.flipkart.flux.task.redriver.RedriverRegistry;
 import com.google.common.collect.Sets;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
 
 import javax.inject.Inject;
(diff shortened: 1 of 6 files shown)


runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -211,7 +212,9 @@ private void executeStates(StateMachine stateMachine, Set<State> executableState
      * @param executableStates states whose all dependencies are met
      */
     private void executeStates(StateMachine stateMachine, Set<State> executableStates, Event currentEvent) {
+        MDC.put("stateMachineId", stateMachine.getId().toString());
```
shyamakirala: keep stateMachineId and taskId as string constants in commons module Constants file

task/src/main/java/com/flipkart/flux/impl/task/AkkaTask.java
```diff
@@ -60,7 +62,8 @@
     /**
      * Logger instance for this class
      */
-    private LoggingAdapter logger = Logging.getLogger(getContext().system(), this);
+    //private LoggingAdapter logger = Logging.getLogger(getContext().system(), this);
```
shyamakirala: Remove the commented out line.

PR #65, opened 2017-05-08
Support for workflow cancellation

shyamakirala:
On calling cancel API, all initialized tasks of the workflow and workflow status would be marked as cancelled.

If any event occurs if a workflow is cancelled state, the event would be discarded.

### DIFF
--- a/runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
+++ b/runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
@@ -148,7 +148,7 @@ public class StateMachineResource {
 
         // 1. Convert to StateMachine (domain object) and save in DB
         StateMachine stateMachine = stateMachinePersistenceService.createStateMachine(stateMachineDefinition);
-        MDC.put(STATE_MACHINE_ID, stateMachine.getId().toString());
+        MDC.clear(); MDC.put(STATE_MACHINE_ID, stateMachine.getId().toString());
         logger.info("Created state machine with Id: {} and correlation Id: {}", stateMachine.getId(), stateMachine.getCorrelationId());
 
         // 2. initialize and start State Machine
@@ -172,7 +172,7 @@ public class StateMachineResource {
                                 @QueryParam("triggerTime") Long triggerTime,
                                 EventData eventData
     ) throws Exception {
-        MDC.put(STATE_MACHINE_ID, machineId);
+        MDC.clear(); MDC.put(STATE_MACHINE_ID, machineId);
 
         if(triggerTime == null) {
             logger.info("Received event: {} for state machine: {}", eventData.getName(), machineId);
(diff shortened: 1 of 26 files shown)


runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -148,7 +147,7 @@ protected StateMachine createAndInitStateMachine(StateMachineDefinition stateMac
 
         // 1. Convert to StateMachine (domain object) and save in DB
         StateMachine stateMachine = stateMachinePersistenceService.createStateMachine(stateMachineDefinition);
-        MDC.put(STATE_MACHINE_ID, stateMachine.getId().toString());
+        MDC.clear(); MDC.put(STATE_MACHINE_ID, stateMachine.getId().toString());
```
amitkumar94: formatting .. should be two separate lines

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -203,7 +202,7 @@ public Response submitEvent(@PathParam("machineId") String machineId,
     ) throws Exception {
         EventData eventData = eventAndExecutionData.getEventData();
         ExecutionUpdateData executionUpdateData = eventAndExecutionData.getExecutionUpdateData();
-        MDC.put(STATE_MACHINE_ID, machineId);
+        MDC.clear(); MDC.put(STATE_MACHINE_ID, machineId);
```
amitkumar94: same as above comment

task/src/main/java/com/flipkart/flux/impl/redriver/RedriverRegistryImpl.java
```diff
@@ -102,7 +102,7 @@ public void registerTask(Long taskId, long redriveDelay) {
 	 * @see RedriverRegistry#deRegisterTask(java.lang.Long)
 	 */
 	public void deRegisterTask(Long taskId) {
-		MDC.put(TASK_ID,taskId.toString());
+		MDC.clear(); MDC.put(TASK_ID,taskId.toString());
```
amitkumar94: same formatting issue

task/src/main/java/com/flipkart/flux/impl/redriver/RedriverRegistryImpl.java
```diff
@@ -112,7 +112,7 @@ public void deRegisterTask(Long taskId) {
 	 * @see RedriverRegistry#redriveTask(java.lang.Long)
 	 */
 	public void redriveTask(Long taskId) {
-		MDC.put(TASK_ID, taskId.toString());
+		MDC.clear(); MDC.put(TASK_ID, taskId.toString());
```
amitkumar94: same

PR #72, opened 2017-12-12
Async Redrive Call changes

amitkumar94:


scheduler/src/main/java/com/flipkart/flux/redriver/service/RedriverService.java
```diff
@@ -103,11 +103,26 @@ private void redrive() {
...
-                } catch (Exception ex) {}
+                    messageRedrived.add(asyncRedriveService.submit(() -> {
+                        redriverRegistry.redriveTask(e.getTaskId());
+                    }));
+                } catch (Exception ex) {
+                }
             });
+            boolean allCompleted = false;
+            while (!allCompleted) {
+                allCompleted = true;
```
shyamakirala: I think it's good to have a small sleep here, else this loop will do many iterations unnecessarily during http call duration.

PR #75, opened 2017-12-29
Configurable Event Proxy Changes for Migration

amitkumar94:


client/pom.xml
```diff
@@ -7,7 +7,7 @@
     <parent>
         <groupId>com.flipkart</groupId>
         <artifactId>flux</artifactId>
-        <version>1.1.4-SNAPSHOT</version>
+        <version>1.2.0-SNAPSHOT</version>
```
shyamakirala: Not part of this one, but delete the earlier 1.2.0-SNAPSHOT git tag, create again as the old code results in compilation errors

runtime/src/main/resources/packaged/configuration.yml
```diff
@@ -79,5 +80,9 @@ eventScheduler:
     batchSize: 50
     intervalms: 10000
 
+eventProxyForMigration:
+  enabled: "yes"
+  endpoint: "http://10.85.42.144"
```
shyamakirala: avoid committing IPs

amitkumar94: was needed for the new tests to pass, will remove now.

client/pom.xml
amitkumar94: still not merging in master, will create new tag from master, with correct version details.

PR #77, opened 2018-05-03
Isolated execution node changes

amitkumar94:


common/src/main/java/com/flipkart/flux/FluxRole.java
```diff
@@ -0,0 +1,7 @@
+package com.flipkart.flux;
+
+public enum FluxRole {
```
ykameshrao: minor: Should we call this FluxRuntimeRole indicating the role is regarding the runtime env?

examples/src/main/java/com/flipkart/flux/examples/benchmark/RunBenchMarkWorkFlowA.java
```diff
@@ -19,7 +19,7 @@ public static void main(String args[]) {
 
         /* Lets invoke our workflow */
         System.out.println("[Main] Starting BenchMarkWorkFlowA  execution");
-        for (int i = 1; i <= 100 ; i++) {
+        for (int i = 1; i <= 1 ; i++) {
```
ykameshrao: why 1? Why is this reduced to 1?

common/src/test/java/com/flipkart/flux/runner/GuiceJunit4Runner.java
```diff
@@ -55,37 +66,73 @@ public GuiceJunit4Runner(Class<?> klass) throws InitializationError, IOException
...
-    }
-
-    private Polyguice getPolyInstance(Object testInstance) {
         final Modules moduleAnnotation = testInstance.getClass().getAnnotation(Modules.class);
         if (moduleAnnotation == null) {
             throw new RuntimeException("You now have to supply a @Modules annotation to use GuiceJunit4Runner");
         }
-        final Class<? extends AbstractModule>[] configuredModules = moduleAnnotation.value();
+        if (moduleAnnotation.executionModules().length > 0) {
+            configModuleForExecution = new ConfigModule(FluxRole.EXECUTION);
```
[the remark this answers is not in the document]
ykameshrao: Actually, I read the code in the local copy and realised that InjectFromRole is actually happening only in test cases and test code. Ignore this.

runtime/src/main/java/com/flipkart/flux/resource/ClientElbResource.java
```diff
@@ -0,0 +1,175 @@
...
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.UUID;
+
+/**
+ * API for CRUD operations on Client Cluster ELB details
+ * @author akif.khan
+ */
+@Singleton
+@Path("/api/clientelb")
```
ykameshrao: /api/client-elb is probably a better name.

runtime/src/main/java/com/flipkart/flux/resource/ClientElbResource.java
```diff
@@ -0,0 +1,175 @@
...
+                                @QueryParam("clientElbUrl") String clientElbUrl) {
+        if(clientElbUrl == null)
+            return Response.status(Response.Status.BAD_REQUEST.getStatusCode()).entity(
+                    "ClientElbUrl cannot be null").build();
+        else {
+            try {
+                URL verifyingURL = new URL(clientElbUrl);
+                verifyingURL.toURI();
+                if(clientId == null)
+                    clientId = UUID.randomUUID().toString();
```
ykameshrao: why is this random? client-id should be mandatory param for the API.

runtime/src/main/java/com/flipkart/flux/representation/ClientElbPersistenceService.java
```diff
@@ -0,0 +1,277 @@
...
+     * @param clientId
+     * @param clientElb
+     */
+    private synchronized void clientElbCacheRefresh(String clientId, ClientElb clientElb) {
+
+        try {
+
+            if(clientElbCache.size() < MAX_CACHE_SIZE) {
+                clientElbCache.put(clientId, clientElb);
+                lastAccessed.put(clientId, System.currentTimeMillis());
```
ykameshrao: why not use GuavaCache or something like that instead of reinventing the wheel?

scheduler/src/main/java/com/flipkart/flux/clientelb/dao/ClientElbDAOImpl.java
```diff
@@ -0,0 +1,135 @@
...
+                .setMaxResults(rowCount)
+                .list();
+    }
+
+    /**
+     * Saves the object in DB and returns the saved object.
+     * @param clientElb
+     * @return saved clientElb object
+     */
+    public ClientElb save(ClientElb clientElb) {
```
ykameshrao: Transactional?

akifkhan01: Removed this method and moved it's content to ONLY calling method which is Transactional:
     /**
     * Creates client elb entry in DB and returns saved object
     */
    @Override
    @Transactional
    @SelectDataSource(storage = Storage.SCHEDULER)
    public ClientElb create(String clientElbId, ClientElb clientElb) {
        return currentSession().save(clientElb);
    }

api/src/main/java/com/flipkart/flux/api/ClientElbDefinition.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /** Constructor*/
+    public ClientElbDefinition(String id, String elbUrl) {
+        super();
+        this.id = id;
+        this.elbUrl = elbUrl;
+    }
+
+    /** Accessors/Mutators for member variables*/
```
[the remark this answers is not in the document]
akifkhan01: No, we haven't used lombok so far, though we should add it going forward.

persistence-mysql/src/main/resources/flux_scheduler/migrations/0003_create_ClientElb_table.sql
```diff
@@ -0,0 +1,17 @@
+--liquibase formatted sql
+
+--changeset akif.khan:3 runOnChange:false
+
+CREATE TABLE IF NOT EXISTS `ClientElb` (
+  `id` VARCHAR(64) NOT NULL,
+  `elbUrl` VARCHAR(100) NOT NULL,
```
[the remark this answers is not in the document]
akifkhan01: Changed VARCHAR length to 400 for 'elbUrl'

runtime/src/main/java/com/flipkart/flux/resource/ClientElbResource.java
akifkhan01: Made client-id field mandatory, no more randomly generated clientIds' .

persistence-mysql/src/main/resources/flux_scheduler/migrations/0003_create_ClientElb_table.sql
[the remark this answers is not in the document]
akifkhan01: Changed to 300

runtime/src/main/java/com/flipkart/flux/resource/ClientElbResource.java
akifkhan01: Changed endpoint to '/api/client-elb' here and at corresponding usage places.

Changes can be viewed in commit titled: Review Comments Resolution 2
Added a new commit because rebase was causing much more merge conflicts
(prone to error and subject to verification) and it was time consuming.

runtime/src/main/java/com/flipkart/flux/representation/ClientElbCacheInitializer.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/**
+ * It initializes in-memory ClientElbCache belonging to Singleton
+ * ClientElbPersistenceService object by retrieving oldest entries from database bounded by maximum cache size mentioned.
+ * @author akif.khan
+ */
+@Singleton
+public class ClientElbCacheInitializer {
```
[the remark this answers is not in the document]
akifkhan01: With Guava cache introduced now, this initialiser is no more required, removed.

See Commit title: Replace custmoized LRU cache with Google Guava Cache...

runtime/src/main/java/com/flipkart/flux/representation/ClientElbPersistenceService.java
akifkhan01: Introduced Google Guava Cache.
See Commit title: Replace customized LRU cache with Google Guava Cache...

runtime/src/main/java/com/flipkart/flux/representation/ClientElbPersistenceService.java
```diff
@@ -0,0 +1,277 @@
...
+            }
+
```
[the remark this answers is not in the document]
akifkhan01: Introduced Google Guava Cache.
See Commit title: Replace customized LRU cache with Google Guava Cache...

api/src/main/java/com/flipkart/flux/api/ClientElbDefinition.java
```diff
@@ -0,0 +1,81 @@
...
+    /** Client Elb URL*/
+    private String elbUrl;
+
+    /* To be used only by jackson */
+    ClientElbDefinition() {
+    }
+
+    /** Constructor*/
+    public ClientElbDefinition(String id, String elbUrl) {
+        super();
```
amitkumar94: why is super required here ? There is no parent class.

api/src/main/java/com/flipkart/flux/api/StateMachineDefinition.java
```diff
@@ -44,19 +44,25 @@
     /* User supplied string for easy identification of a workflow instance */
     private String correlationId;
 
+    /* Client Eb Id to which task is supposed to be forwarded for execution in Flux Runtime  */
+    private String clientElbId;
```
amitkumar94: Shouldn't this and correlationId field be a part of hascode, toString and Equals methods ?

client/src/main/java/com/flipkart/flux/client/config/FluxClientConfiguration.java
```diff
@@ -31,16 +31,20 @@
 public class FluxClientConfiguration {
 
     private String fluxRuntimeUrl = "http://localhost:9998";
+    private String clientElbId = "client_elb_id_1";
```
amitkumar94: Is this the default clientElbId ? If it is use a better name like "defaultElb" or somethin, which should be a global constant as well.

client/src/main/java/com/flipkart/flux/client/config/FluxClientConfiguration.java
```diff
@@ -62,4 +66,11 @@ public long getConnectionTimeout() {
     public void setConnectionTimeout(long connectionTimeout) {
         this.connectionTimeout = connectionTimeout;
     }
+    public String getClientElbId() {
+        return clientElbId;
+    }
+    public void setClientElbId(String clientElbId) {
```
amitkumar94: Also this field should be immutable.

client/src/test/java/com/flipkart/flux/client/runtime/LocalContextTest.java
```diff
@@ -43,23 +43,29 @@ public void setUp() throws Exception {
 
     @Test
     public void testRegisterNew_shouldCreateALocalRegistration() throws Exception {
-        localContext.registerNew("fooBar", 1, "someDescription","someContextId");
-        assertThat(tlStateMachineDef.get()).isEqualTo(new StateMachineDefinition("someDescription", "fooBar", 1l, new HashSet<>(), new HashSet<>(), "someContextId"));
+        localContext.registerNew("fooBar", 1, "someDescription",
```
amitkumar94: As discussed earlier, declare this constant globally in client Module, on another note are we allowing null for clientElb ?

common/src/test/java/com/flipkart/flux/runner/GuiceJunit4Runner.java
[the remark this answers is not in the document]
amitkumar94: This was added to dynamically inject only required guice modules to test classes at runtime. All the modules are prepared in two polyInstances only once and re used wherever applicable.

common/src/test/java/com/flipkart/flux/runner/Modules.java
```diff
@@ -26,5 +26,6 @@
 @Target(ElementType.TYPE)
 @Documented
 public @interface Modules {
```
amitkumar94: self comment, update java doc for this.

common/src/test/java/com/flipkart/flux/runner/Modules.java
```diff
@@ -26,5 +26,6 @@
...
-    Class<? extends AbstractModule>[] value();
+    Class<? extends AbstractModule>[] orchestrationModules() default {};
```
amitkumar94: Update javadoc

model/src/main/java/com/flipkart/flux/domain/ClientElb.java
```diff
@@ -0,0 +1,116 @@
...
+ * limitations under the License.
+ */
+
+package com.flipkart.flux.domain;
+
+import javax.persistence.*;
+import java.sql.Timestamp;
+
+/**
+ * <code>ClientElb</code> represents Client Cluster's Elb details.
```
amitkumar94: Elaborate more in java doc, for eg Entity which contains info regarding the endpoint of a clients execution environment, maintained in mysql.

model/src/main/java/com/flipkart/flux/domain/ClientElb.java
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Constructors
+     */
+    protected ClientElb() {
+    }
+
+    public ClientElb(String id, String elbUrl) {
+        super();
+        this.id = id;
+        this.elbUrl = elbUrl;
```
amitkumar94: Have a static helper method to validate the elbUrl.

model/src/main/java/com/flipkart/flux/domain/ClientElb.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public String getElbUrl() {
+        return elbUrl;
+    }
+
+    public void setElbUrl(String elbUrl) {
+        this.elbUrl = elbUrl;
+    }
+
+    public Timestamp getCreatedAt() {
```
amitkumar94: Accessor/Mutator  method not required for timestamps.

model/src/main/java/com/flipkart/flux/domain/ClientElb.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof ClientElb)) return false;
+
+        ClientElb that = (ClientElb) o;
+
+        if (createdAt != null ? !createdAt.equals(that.createdAt) : that.createdAt != null) return false;
+        if (elbUrl != null ? !elbUrl.equals(that.elbUrl) : that.elbUrl != null) return false;
```
amitkumar94: same, createdAt/updatedAt should not be part of equals, hashcode, toString

model/src/main/java/com/flipkart/flux/domain/StateMachine.java
```diff
@@ -158,6 +166,14 @@ public Timestamp getUpdatedAt() {
         return updatedAt;
     }
 
+    public String getClientElbId() {
+        return clientElbId;
+    }
+
+    public void setClientElbId(String clientElbId) {
```
amitkumar94: no setter, to be initialized one time in constructor

persistence-mysql/src/main/resources/flux_scheduler/migrations/0003_create_ClientElb_table.sql
```diff
@@ -0,0 +1,17 @@
+--liquibase formatted sql
+
+--changeset akif.khan:3 runOnChange:false
+
+CREATE TABLE IF NOT EXISTS `ClientElb` (
+  `id` VARCHAR(64) NOT NULL,
+  `elbUrl` VARCHAR(300) NOT NULL,
+  `createdAt` TIMESTAMP(3) DEFAULT CURRENT_TIMESTAMP(3),
```
amitkumar94: createdAt` TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),

persistence-mysql/src/main/resources/flux_scheduler/migrations/0003_create_ClientElb_table.sql
```diff
@@ -0,0 +1,17 @@
...
+  `updatedAt` TIMESTAMP(3) DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3),
```
amitkumar94: `updatedAt` TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3),

runtime/pom.xml
```diff
@@ -283,6 +283,13 @@
             <version>${codahale.metrics.version}</version>
         </dependency>
 
+        <!-- In-Memory Guava Cache dependencies-->
```
amitkumar94: This is not required.

runtime/src/main/java/com/flipkart/flux/controller/FSMVisualizationController.java
```diff
@@ -39,19 +39,13 @@
...
-        this.configInjector = configInjector;
-    }
 
     @RequestMapping(value = {"/fsmview"}, method = RequestMethod.GET)
     public String fsmview(@RequestParam(value = "fsmid", required = false) String fsmId, ModelMap modelMap, HttpServletRequest request) throws UnknownHostException {
-        int fluxApiPort = configInjector.getInstance(Key.get(Integer.class, Names.named("Api.service.port")));
+//        int fluxApiPort = configInjector.getInstance(Key.get(Integer.class, Names.named("Api.service.port")));
         String fluxApiHost = InetAddress.getLocalHost().getHostAddress();
-        String fluxApiUrl = System.getProperty("flux.runtimeUrl")== null ? "http://" + fluxApiHost   + ":" + fluxApiPort : System.getProperty("flux.runtimeUrl");
+        String fluxApiUrl = System.getProperty("flux.runtimeUrl")== null ? "http://" + fluxApiHost   + ":" + 9998 : System.getProperty("flux.runtimeUrl");
```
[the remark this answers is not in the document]
amitkumar94: cause the dashboard server and api server run on different port, and dashboard server makes a call to api server to get fsm related details. will see if this can be refactored better.

runtime/src/main/java/com/flipkart/flux/representation/ClientElbPersistenceService.java
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * Persists the ClientElb details object in DB.
+     *
+     * @param clientElbDefinition
+     * @return created clientElb
+     */
+    @Transactional
+    @SelectDataSource(storage = Storage.SCHEDULER)
+    public ClientElb persistClientElb(String clientId, ClientElbDefinition clientElbDefinition) {
```
amitkumar94: Why do these functions need to be transactional, if the underlying functions are transactions already.  Also these functions should be part of the api layer which accesss DAO .

runtime/src/main/java/com/flipkart/flux/representation/ClientElbPersistenceService.java
```diff
@@ -0,0 +1,159 @@
...
+ * @author akif.khan
+ */
+@Singleton
+public class ClientElbPersistenceService {
+
+    private ClientElbDAO clientElbDAO;
+
+    private LoadingCache<String, ClientElb> clientElbCache;
+
+    private Integer MAX_CACHE_SIZE;
```
amitkumar94: Why is this cache limited size ?

runtime/src/main/java/com/flipkart/flux/representation/ClientElbPersistenceService.java
```diff
@@ -0,0 +1,159 @@
...
+    public boolean clientElbCacheContainsKey(String key) {
+        return this.clientElbCache.asMap().containsKey(key);
+    }
+
+    public String getClientElbCacheUrl(String id) {
+        String clientElbUrl = null;
+        try {
+            clientElbUrl = this.clientElbCache.get(id).getElbUrl();
+        } catch (ExecutionException e) {
+            logger.error("ClientElbCache entry not found in both DB and cache {} {}", e.getMessage(), e.getStackTrace());
```
amitkumar94: where is the check to db made ?

runtime/src/main/java/com/flipkart/flux/representation/ClientElbPersistenceService.java
```diff
@@ -0,0 +1,159 @@
...
+ * an in-memory cache(Google Guava: https://github.com/google/guava/wiki) of clientElb entries for faster query results.
+ *
+ * @author akif.khan
+ */
+@Singleton
+public class ClientElbPersistenceService {
+
+    private ClientElbDAO clientElbDAO;
+
+    private LoadingCache<String, ClientElb> clientElbCache;
```
amitkumar94: isnt the key same as clientElb.elbId  and value clientElb.url ?  Can't we just store a cache of <String, String>

runtime/src/main/resources/packaged/test_config.yml
```diff
@@ -0,0 +1,103 @@
...
+
+flux_scheduler.Hibernate:
+  hibernate.dialect: "org.hibernate.dialect.MySQLDialect"
+  hibernate.connection.driver_class: "com.mysql.jdbc.Driver"
+  hibernate.connection.url: "jdbc:mysql://localhost:3306/flux_scheduler"
+  hibernate.connection.username: "root"
+  hibernate.connection.password: ""
+  hibernate.current_session_context_class: "managed"
+  connection.provider_class: "org.hibernate.c3p0.internal.C3P0ConnectionProvider"
+  hibernate.c3p0.min_size: 3
```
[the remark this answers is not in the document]
amitkumar94: These configurations are only for local testing.

runtime/src/test/java/com/flipkart/flux/integration/E2ETest.java
```diff
@@ -14,76 +14,105 @@
...
-    ExecutableRegistry registry;
+    @InjectFromRole(value = FluxRole.EXECUTION)
+    TaskExecutableRegistryImpl registry;
+
+    @InjectFromRole(value = FluxRole.EXECUTION)
+    ExecutionOrderedComponentBooter executionOrderedComponentBooter;
+
+    @Before
+    public void setUp() {
+        try {
```
amitkumar94: What is the purpose of this ?  Shouldn't these api call be covered as test in separate class.

runtime/src/test/java/com/flipkart/flux/representation/ClientElbPersistenceServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+        String elbUrl3 = "http://10.4.3.3";
+        clientElbDAO.create(clientId3, new ClientElb(clientId3, elbUrl3));
+
+        String clientId4 = "id4";
+        String elbUrl4 = "http://10.4.3.3";
+        clientElbDAO.create(clientId4, new ClientElb(clientId4, elbUrl4));
+    }
+
+    @Test
+    public void testClientElbCacheRefresherLRUPolicy() {
```
amitkumar94: We don't need a refreshing policy on cache, it is not that big, will hardly have 10 entries.

runtime/src/test/java/com/flipkart/flux/resource/ClientElbResourceTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    ClientElbPersistenceService clientElbPersistenceService = mock(ClientElbPersistenceService.class);
+
+    ClientElbResource clientElbResource = new ClientElbResource(clientElbPersistenceService);
+
+    @Test
+    public void testCreateClientElb() {
+
+        ClientElbDefinition clientElbDefinition = new ClientElbDefinition(
+                "id1", "http://10.3.2.3/api/resource/");
```
amitkumar94: Client Elb can't contain path in the url, it should be just plain elb address. Add this validation to the base class as well.

runtime/src/test/java/com/flipkart/flux/resource/ClientElbResourceTest.java
```diff
@@ -0,0 +1,113 @@
...
+        verify(clientElbPersistenceService, times(1)).persistClientElb(
+                "id1", clientElbDefinition);
+        verifyNoMoreInteractions(clientElbPersistenceService);
+    }
+
+
+    @Test
+    public void testFindById() {
+
+        ClientElb clientElb = new ClientElb("id2", "http://10.3.2.3/api");
```
amitkumar94: create a helper method to do this, avoid duplicacy.

runtime/src/test/java/com/flipkart/flux/resource/StateMachineResourceTest.java
```diff
@@ -92,15 +100,30 @@
     public void setUp() throws Exception {
         objectMapper = new ObjectMapper();
         dbClearRule.explicitClearTables();
+    }
 
+    @BeforeClass
+    public static void beforeClass() {
+        try {
```
amitkumar94: If the same setup is used multiple places, extract to helper method and use that.

scheduler/src/main/java/com/flipkart/flux/clientelb/dao/ClientElbDAOImpl.java
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Updates Elb Url of a client in DB
+     */
+    @Override
+    @Transactional
+    @SelectDataSource(storage = Storage.SCHEDULER)
+    public void updateElbUrl(String clientElbId, String clientElbUrl) {
+        Query updateQuery = currentSession().createQuery(
+                "update ClientElb set elbUrl = :clientElbUrl where id = :clientElbId");
+        updateQuery.setString("clientElbUrl", clientElbUrl != null ? clientElbUrl.toString() : null);
```
amitkumar94: Null clientElbUrl should never reach here, those calls should be reject on api layer itself.

scheduler/src/main/java/com/flipkart/flux/clientelb/dao/ClientElbDAOImpl.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * Retrieve Older Client Elb entry limit by parameter rowCount
+     * @param rowCount
+     * @return List of oldest ClientElbs'
+     */
+    @Override
+    @Transactional
+    @SelectDataSource(storage = Storage.SCHEDULER)
```
amitkumar94: We don't need refreshing policy.

scheduler/src/main/java/com/flipkart/flux/clientelb/dao/ClientElbDAOImpl.java
```diff
@@ -0,0 +1,126 @@
...
+        this.sessionFactoryContext = sessionFactoryContext;
+    }
+
+    /**
+     * Creates client elb entry in DB and returns saved object
+     */
+    @Override
+    @Transactional
+    @SelectDataSource(storage = Storage.SCHEDULER)
+    public ClientElb create(String clientElbId, ClientElb clientElb) {
```
amitkumar94: Two arguments are not required, as first one can be extracted from the second one itself.

task/src/main/java/com/flipkart/flux/api/core/TaskExecutionMessage.java
```diff
@@ -0,0 +1,66 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.flipkart.flux.api.core;
+
+import com.flipkart.flux.impl.message.TaskAndEvents;
+
```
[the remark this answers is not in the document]
amitkumar94: We are not storing this object in any store, so serialisation not required, Object is created on need basis and sent via http for execution.

task/src/main/java/com/flipkart/flux/taskDispatcher/TaskDispatcher.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.flipkart.flux.taskDispatcher;
+
+import com.flipkart.flux.api.core.TaskExecutionMessage;
+
+/**
+ * Interface to forward TaskExecutionMessages to remote Execution nodes.
+ */
+public interface TaskDispatcher {
```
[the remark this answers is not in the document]
amitkumar94: We can have a generic dispatcher for http calls made by orchestration nodes, which could support multiple methods in future for different implementation.

task/src/main/java/com/flipkart/flux/api/core/TaskExecutionMessage.java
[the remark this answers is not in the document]
amitkumar94: Also the object is sent as byte Array via http layer.

model/src/main/java/com/flipkart/flux/domain/ClientElb.java
```diff
@@ -0,0 +1,116 @@
...
+    public ClientElb(String id, String elbUrl) {
+        super();
+        this.id = id;
+        this.elbUrl = elbUrl;
+    }
+
+    /**
+     * Accessor/Mutator methods
+     */
+    public String getId() {
```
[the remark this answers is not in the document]
akifkhan01: We can add lombok to our complete DAO layer itself in a separate re-factoring exercise.

client/src/main/java/com/flipkart/flux/client/config/FluxClientConfiguration.java
akifkhan01: Added "defaultElbId" as a default name.

model/src/main/java/com/flipkart/flux/domain/ClientElb.java
akifkhan01: We are validating the URL at API resource layer itself.

model/src/main/java/com/flipkart/flux/domain/ClientElb.java
akifkhan01: Removed Accessor/Mutator methods for Timestamps createdAt and updatedAt.

model/src/main/java/com/flipkart/flux/domain/ClientElb.java
akifkhan01: Removed from equals, hashcode and toString methods.

client/src/main/java/com/flipkart/flux/client/config/FluxClientConfiguration.java
akifkhan01: Removed setter for ClientElbId, immutable now.

model/src/main/java/com/flipkart/flux/domain/StateMachine.java
akifkhan01: Removed setter for ClientElbId.

persistence-mysql/src/main/resources/flux_scheduler/migrations/0003_create_ClientElb_table.sql
akifkhan01: Made this Timestamp field as NOT NULL.

persistence-mysql/src/main/resources/flux_scheduler/migrations/0003_create_ClientElb_table.sql
akifkhan01: Made this Timestamp field as NOT NULL.

runtime/pom.xml
akifkhan01: Removed maven dependency. It's already added as part of com.google.common.cache package
dependency.

runtime/src/main/java/com/flipkart/flux/representation/ClientElbPersistenceService.java
akifkhan01: Removed @Transactional from these Service layer CRUD methods. Earlier, before introducing
Guava Cache, I defined a custom LRU cache where these methods were @Transactional, not required anymore. 
findById, update and delete need to be a part of this cache service layer because in each of these cases, cache needs to read, refresh/update and delete entry respectively.
We can afford to move create operation directly to API layer, however for the modularity of the code, I think we should let it pass via this cache service layer itself.

runtime/src/main/java/com/flipkart/flux/representation/ClientElbPersistenceService.java
akifkhan01: We are passing the cache size via config file. Though, in our use case it is not required but 
it is a good practice to keep the cache size configurable which makes the testing of cache
operations easier.

runtime/src/main/java/com/flipkart/flux/representation/ClientElbPersistenceService.java
akifkhan01: In case of cache miss, LoadingCache makes a call to @override method "load" where we make a DB call to fetch results. Added check for returned "null" value.

runtime/src/main/java/com/flipkart/flux/representation/ClientElbPersistenceService.java
akifkhan01: Changed Cache to <String, String> from <String, ClientElb>.

runtime/src/test/java/com/flipkart/flux/representation/ClientElbPersistenceServiceTest.java
akifkhan01: Removed the test.

scheduler/src/main/java/com/flipkart/flux/clientelb/dao/ClientElbDAOImpl.java
akifkhan01: Removed redundant NULL check, we are already checking for ClientElbUrl NULL value at API layer itself for both create and update operations.

scheduler/src/main/java/com/flipkart/flux/clientelb/dao/ClientElbDAOImpl.java
akifkhan01: Changed method signature to 
public ClientElb create(ClientElb clientElb) {

client/src/main/java/com/flipkart/flux/client/config/FluxClientConfiguration.java
akifkhan01: Removed setter for ClientELbId, it's immutable now.

model/src/main/java/com/flipkart/flux/domain/ClientElb.java
akifkhan01: Added more elaborative javadoc about ClientElb Entity.

scheduler/src/main/java/com/flipkart/flux/clientelb/dao/ClientElbDAOImpl.java
akifkhan01: Added this retrieveOldest to retrieve oldest entry sorted by creation time upto a specified count to support LRU policy in customized cache.
Since, we are using now Guava Cache, don't need this function anymore, removing it.
Also, as per our use case, cache won't need rotation policy.

runtime/src/main/java/com/flipkart/flux/representation/ClientElbPersistenceService.java
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * @param clientId
+     * @param clientElbUrl
+     */
+    @Transactional
+    @SelectDataSource(storage = Storage.SCHEDULER)
+    public void updateClientElb(String clientId, String clientElbUrl) {
+        clientElbDAO.updateElbUrl(clientId, clientElbUrl);
+        synchronized (this) {
+            this.clientElbCache.refresh(clientId);
```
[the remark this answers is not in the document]
akifkhan01: Added Synchronised block here because retrieval operations (including get) generally do not block, so may overlap with update operations (including put, refresh and remove). Retrievals reflect the results of the most recently completed update operations holding upon their onset.
However, since in our use case, call to update/delete would be very less frequent and additionally, even if a dirty read occurs, call will get fixed in second attempt as soon as entry in cache is updated/deleted.
Removing synchronised block for both update and delete.

client/src/test/java/com/flipkart/flux/client/runtime/LocalContextTest.java
akifkhan01: I have added it's default value in FluxClientConfiguration. For other usages, it is being used
via config files wherever possible.
No, we don't allow NULL value for ClientElbId in StateMachines table, if not specified it takes
'defaultElbId' as default value.

runtime/src/test/java/com/flipkart/flux/resource/ClientElbResourceTest.java
akifkhan01: Added check using java.net.URL.getPath() method which checks whether path exists in input
URL.

common/src/main/java/com/flipkart/flux/guice/module/ConfigModule.java
```diff
@@ -53,6 +67,10 @@ public ConfigModule() {
         }
     }
 
+    public ConfigModule() {
```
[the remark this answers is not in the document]
amitkumar94: no longer required this was supposed to be invoked at runtime via spring to get a config key for port, currently hardcoding the port

task/src/main/java/com/flipkart/flux/taskDispatcher/TaskDispatcherImpl.java
```diff
@@ -0,0 +1,106 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+@Singleton
+public class TaskDispatcherImpl implements TaskDispatcher {
+
+    private static Logger logger = LoggerFactory.getLogger(TaskDispatcherImpl.class);
+    //Default , should be overridden via config if necessary
+    // Need to add a metric for this api as well
+    public static final int MAX_TOTAL_CONNECTIONS = 100;
```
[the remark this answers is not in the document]
amitkumar94: done

task/src/main/java/com/flipkart/flux/taskDispatcher/TaskDispatcherImpl.java
```diff
@@ -0,0 +1,106 @@
...
+                .setConnectionRequestTimeout((TaskDispatcherImpl.socketTimeOut).intValue())
+                .build();
+        PoolingHttpClientConnectionManager syncConnectionManager = new PoolingHttpClientConnectionManager();
+        syncConnectionManager.setMaxTotal(MAX_TOTAL_CONNECTIONS);
+        syncConnectionManager.setDefaultMaxPerRoute(MAX_CONNECTIONS_PER_ROUTE);
+        closeableHttpClient = HttpClientBuilder.create().setDefaultRequestConfig(clientConfig).setConnectionManager(syncConnectionManager)
+                .build();
+
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            HttpClientUtils.closeQuietly(closeableHttpClient);
```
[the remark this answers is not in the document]
amitkumar94: This can be taken as when refactoring in future.

task/src/main/java/com/flipkart/flux/taskDispatcher/TaskDispatcherImpl.java
```diff
@@ -0,0 +1,106 @@
...
+
+
+    @Override
+    public int forwardExecutionMessage(String endPoint, TaskExecutionMessage taskExecutionMessage) {
+        int defaultStatusCode = -1;
+        CloseableHttpResponse httpResponse = null;
+        HttpPost httpPostRequest;
+        httpPostRequest = new HttpPost(endPoint);
+        try {
+            final ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
```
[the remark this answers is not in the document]
amitkumar94: Not required as it garbage collected.
Closing a ByteArrayOutputStream has no effect. The methods in this class can be called after the stream has been closed without generating an IOException.

scheduler/src/test/java/com/flipkart/flux/eventscheduler/dao/EventSchedulerDaoTest.java
```diff
@@ -35,14 +37,13 @@
  * @author shyam.akirala
  */
 @RunWith(GuiceJunit4Runner.class)
-@Modules({ConfigModule.class, SchedulerTestModule.class})
+@Modules(orchestrationModules = {SchedulerTestModule.class}, executionModules = {})
 public class EventSchedulerDaoTest {
 
-    @Inject
+    @InjectFromRole(value = FluxRole.ORCHESTRATION)
```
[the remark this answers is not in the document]
amitkumar94: nope, its the other way around, dependency is to be injected from an already prepared instance by guice. This annotation is only valid for test classes only.

examples/src/main/java/com/flipkart/flux/examples/benchmark/RunBenchMarkWorkFlowA.java
amitkumar94: This is a helper class to generate few workflows at local.

scheduler/src/main/java/com/flipkart/flux/redriver/dao/MessageDao.java
```diff
@@ -80,10 +84,13 @@ public void deleteInBatch(List<SmIdAndTaskIdPair> messageIdsToDelete) {
                     .append(smIdAndTaskIdPair.getSmId())
                     .append("\',\'").append(smIdAndTaskIdPair.getTaskId())
                     .append("\'),");
+            logger.info(smIdAndTaskIdPair.toString() + ",");
```
akifkhan01: We don't need this logging info, query is being printed in line 90.

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -483,26 +489,38 @@ private void executeStates(StateMachine stateMachine, Set<State> executableState
                     if (state.getStatus().equals(Status.initialized) || state.getStatus().equals(Status.unsidelined)) {
                         msg.setFirstTimeExecution(true);
                     }
+
                     // register the Task with the redriver
-                    if (state.getRetryCount() > 0) {
```
akifkhan01: Why is this check removed ?

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -493,10 +496,16 @@ private void executeStates(StateMachine stateMachine, Set<State> executableState
                     // register the Task with the redriver
                     // Delay between retires is exponential (2, 4, 8, 16, 32.... seconds) as seen in AkkaTask.
                     // Redriver interval is set as 2 x ( 2^(retryCount+1) x 1s + (retryCount+1) x timeout)
-                    long redriverInterval = 2 * ((int) Math.pow(2, state.getRetryCount() + 1) * 1000 + (state.getRetryCount() + 1) * state.getTimeout());
+                    long redriverInterval;
+                    if (redriverTriggered && state.getStatus() == Status.initialized) {
+                        redriverInterval = 2 * ((int) Math.pow(2, 7) * 1000);
```
akifkhan01: For 512 seconds, it should be:
redriverInterval = 2 * ((int) Math.pow(2, 8) * 1000);

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -493,10 +496,16 @@ private void executeStates(StateMachine stateMachine, Set<State> executableState
...
+                    }
+                    else {
+                        redriverInterval = 2 * ((int) Math.pow(2, state.getRetryCount() + 1) * 1000 + (state.getRetryCount() + 1) * state.getTimeout());
+                    }
                     this.redriverRegistry.registerTask(state.getId(), state.getStateMachineId(), redriverInterval);
 
-                    //send the message to the Router
+                    //send t=dhe message to the Router
```
akifkhan01: Need to correct the typo here.

PR #78, opened 2018-07-16
Event data edit

akifkhan01:
Pull request for Event Data Update Feature.
Refer doc here: https://docs.google.com/document/d/1qpIQ6rskiHK843dOCDVPC-FHXJnayX95QeTU95PBU90/edit

### DIFF
--- a/client/src/main/java/com/flipkart/flux/client/runtime/FluxRuntimeConnectorHttpImpl.java
+++ b/client/src/main/java/com/flipkart/flux/client/runtime/FluxRuntimeConnectorHttpImpl.java
@@ -51,7 +51,6 @@ public class FluxRuntimeConnectorHttpImpl implements FluxRuntimeConnector {
     public static final int MAX_TOTAL = 200;
     public static final int MAX_PER_ROUTE = 20;
     public static final String EXTERNAL = "external";
-    public static final String EXTERNAL_UPDATE = "externalUpdate";
     private final CloseableHttpClient closeableHttpClient;
     private final String fluxEndpoint;
     private final ObjectMapper objectMapper;
@@ -154,7 +153,7 @@ public class FluxRuntimeConnectorHttpImpl implements FluxRuntimeConnector {
     public void submitEventUpdate(String name, Object data, String correlationId,String eventSource) {
         final String eventType = data.getClass().getName();
         if (eventSource == null) {
-            eventSource = EXTERNAL_UPDATE;
+            eventSource = EXTERNAL;
         }
         CloseableHttpResponse httpResponse = null;
         try {
--- a/client/src/main/java/com/flipkart/flux/client/runtime/FluxRuntimeConnectorHttpImpl.java
+++ b/client/src/main/java/com/flipkart/flux/client/runtime/FluxRuntimeConnectorHttpImpl.java
@@ -51,6 +51,7 @@ public class FluxRuntimeConnectorHttpImpl implements FluxRuntimeConnector {
     public static final int MAX_TOTAL = 200;
     public static final int MAX_PER_ROUTE = 20;
     public static final String EXTERNAL = "external";
+    public static final String EXTERNAL_UPDATE = "externalUpdate";
     private final CloseableHttpClient closeableHttpClient;
     private final String fluxEndpoint;
     private final ObjectMapper objectMapper;
@@ -142,6 +143,24 @@ public class FluxRuntimeConnectorHttpImpl implements FluxRuntimeConnector {
                 final EventData eventData = new EventData(name, eventType, (String) data, eventSource);
                 httpResponse = postOverHttp(eventData, "/" + correlationId + "/context/events?searchField=correlationId");
             }
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException(e);
+        } finally {
+            HttpClientUtils.closeQuietly(httpResponse);
+        }
+    }
+
+    @Override
+    public void submitEventUpdate(String name, Object data, String correlationId,String eventSource) {
+        final String eventType = data.getClass().getName();
+        if (eventSource == null) {
+            eventSource = EXTERNAL_UPDATE;
+        }
+        CloseableHttpResponse httpResponse = null;
+        try {
+            final EventData eventData = new EventData(name, eventType, objectMapper.writeValueAsString(data), eventSource);
+            httpResponse = postOverHttp(eventData, "/" + correlationId + "/context/eventupdate");
+
         } catch (JsonProcessingException e) {
             throw new RuntimeException(e);
         } finally {
--- a/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
+++ b/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
@@ -543,7 +543,8 @@ public class WorkFlowExecutionController {
      */
     public void cancelWorkflow(StateMachine stateMachine) {
         stateMachinesDAO.updateStatus(stateMachine.getId(), StateMachineStatus.cancelled);
-        stateMachine.getStates().stream().filter(state -> state.getStatus() == Status.initialized).forEach(state -> {
+        stateMachine.getStates().stream().filter(state -> state.getStatus() == Status.initialized ||
+            state.getStatus() == Status.errored || state.getStatus() == Status.sidelined).forEach(state -> {
             this.statesDAO.updateStatus(stateMachine.getId(), state.getId(), Status.cancelled);
             this.auditDAO.create(stateMachine.getId(), new AuditRecord(stateMachine.getId(), state.getId(), state.getAttemptedNoOfRetries(), Status.cancelled, null, null));
         });
--- a/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
+++ b/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
@@ -42,6 +42,7 @@ import org.slf4j.LoggerFactory;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import javax.transaction.Transactional;
+import javax.ws.rs.core.Response;
 import java.io.IOException;
 import java.util.*;
 
@@ -377,7 +378,7 @@ public class WorkFlowExecutionController {
      * @param stateMachineId
      * @param stateId
      */
-    public void unsidelineState(String stateMachineId, Long stateId) {
+    public Response unsidelineState(String stateMachineId, Long stateId) {
         State askedState = null;
         StateMachine stateMachine = retrieveStateMachine(stateMachineId);
         if (stateMachine == null)
@@ -396,18 +397,16 @@ public class WorkFlowExecutionController {
         Set<State> checkExecutableState = new HashSet<>();
         checkExecutableState.add(askedState);
         if(getExecutableStates(checkExecutableState, stateMachineId).isEmpty()) {
-            logger.error("Cannot unsideline state: {}, at least one dependent event is in pending status.", askedState);
+            logger.error("Cannot unsideline state: {}, at least one of the dependent events is in pending status.", askedState);
+            return Response.status(Response.Status.FORBIDDEN).build();
         }
         else if (askedState.getStatus() == Status.sidelined || askedState.getStatus() == Status.errored) {
             askedState.setStatus(Status.unsidelined);
             askedState.setAttemptedNoOfRetries(0L);
-
             this.statesDAO.updateState(stateMachineId, askedState);
-
             this.executeStates(stateMachine, Sets.newHashSet(Arrays.asList(askedState)));
         }
-
-
+        return Response.status(Response.Status.ACCEPTED).build();
     }
 
     /**
--- a/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
+++ b/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
@@ -42,6 +42,7 @@ import org.slf4j.LoggerFactory;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import javax.transaction.Transactional;
+import javax.ws.rs.ForbiddenException;
 import javax.ws.rs.core.Response;
 import java.io.IOException;
 import java.util.*;
@@ -378,7 +379,7 @@ public class WorkFlowExecutionController {
      * @param stateMachineId
      * @param stateId
      */
-    public Response unsidelineState(String stateMachineId, Long stateId) {
+    public void unsidelineState(String stateMachineId, Long stateId) throws ForbiddenException{
         State askedState = null;
         StateMachine stateMachine = retrieveStateMachine(stateMachineId);
         if (stateMachine == null)
@@ -399,8 +400,7 @@ public class WorkFlowExecutionController {
         if(getExecutableStates(checkExecutableState, stateMachineId).isEmpty()) {
             logger.error("Cannot unsideline state: {}, at least one of the dependent events is in pending status.",
                     askedState.getName());
-            return Response.status(Response.Status.FORBIDDEN.getStatusCode()).entity(
-                    "Cannot unsideline state, at least one of the dependent events is in pending status.").build();
+            throw new ForbiddenException();
         }
         else if (askedState.getStatus() == Status.initialized || askedState.getStatus() == Status.sidelined
                 || askedState.getStatus() == Status.errored) {
@@ -409,7 +409,6 @@ public class WorkFlowExecutionController {
             this.statesDAO.updateState(stateMachineId, askedState);
             this.executeStates(stateMachine, Sets.newHashSet(Arrays.asList(askedState)));
         }
-        return Response.status(Response.Status.ACCEPTED).build();
     }
 
     /**
--- a/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
+++ b/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
@@ -393,7 +393,12 @@ public class WorkFlowExecutionController {
             throw new IllegalStateException("State with the asked id: " + stateId + " not found in stateMachine with id: " + stateMachineId);
         }
 
-        if (askedState.getStatus() == Status.sidelined || askedState.getStatus() == Status.errored) {
+        Set<State> checkExecutableState = new HashSet<>();
+        checkExecutableState.add(askedState);
+        if(getExecutableStates(checkExecutableState, stateMachineId).isEmpty()) {
+            logger.error("Cannot unsideline state: {}, at least one dependent event is in pending status.", askedState);
+        }
+        else if (askedState.getStatus() == Status.sidelined || askedState.getStatus() == Status.errored) {
             askedState.setStatus(Status.unsidelined);
             askedState.setAttemptedNoOfRetries(0L);
 
--- a/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
+++ b/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
@@ -397,8 +397,10 @@ public class WorkFlowExecutionController {
         Set<State> checkExecutableState = new HashSet<>();
         checkExecutableState.add(askedState);
         if(getExecutableStates(checkExecutableState, stateMachineId).isEmpty()) {
-            logger.error("Cannot unsideline state: {}, at least one of the dependent events is in pending status.", askedState);
-            return Response.status(Response.Status.FORBIDDEN).build();
+            logger.error("Cannot unsideline state: {}, at least one of the dependent events is in pending status.",
+                    askedState.getName());
+            return Response.status(Response.Status.FORBIDDEN.getStatusCode()).entity(
+                    "Cannot unsideline state, at least one of the dependent events is in pending status.").build();
         }
         else if (askedState.getStatus() == Status.initialized || askedState.getStatus() == Status.sidelined
                 || askedState.getStatus() == Status.errored) {
--- a/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
+++ b/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
@@ -400,7 +400,8 @@ public class WorkFlowExecutionController {
             logger.error("Cannot unsideline state: {}, at least one of the dependent events is in pending status.", askedState);
             return Response.status(Response.Status.FORBIDDEN).build();
         }
-        else if (askedState.getStatus() == Status.sidelined || askedState.getStatus() == Status.errored) {
+        else if (askedState.getStatus() == Status.initialized || askedState.getStatus() == Status.sidelined
+                || askedState.getStatus() == Status.errored) {
             askedState.setStatus(Status.unsidelined);
             askedState.setAttemptedNoOfRetries(0L);
             this.statesDAO.updateState(stateMachineId, askedState);
--- a/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
+++ b/runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
@@ -379,7 +379,7 @@ public class WorkFlowExecutionController {
      * @param stateMachineId
      * @param stateId
      */
-    public void unsidelineState(String stateMachineId, Long stateId) throws ForbiddenException{
+    public void unsidelineState(String stateMachineId, Long stateId) throws ForbiddenException {
         State askedState = null;
         StateMachine stateMachine = retrieveStateMachine(stateMachineId);
         if (stateMachine == null)
--- a/runtime/src/main/java/com/flipkart/flux/dao/StatesDAOImpl.java
+++ b/runtime/src/main/java/com/flipkart/flux/dao/StatesDAOImpl.java
@@ -131,6 +131,16 @@ public class StatesDAOImpl extends AbstractDAO<State> implements StatesDAO {
         return query.list();
     }
 
-
-
+    @Override
+    @Transactional
+    @SelectDataSource(type = DataSourceType.READ_ONLY, storage = Storage.SHARDED)
+    public List findStatesByDependentEvent(ShardId shardId, String stateMachineId, String eventName) {
+        Query query;
+        String queryString = "select id, stateMachineId, status from State where stateMachineId = :stateMachineId" +
+                " and dependencies like :eventName";
+        query = currentSession().createQuery(queryString);
+        query.setString("stateMachineId", stateMachineId);
+        query.setString("eventName", "%"+eventName+"%");
+        return query.list();
+    }
 }
--- a/runtime/src/main/java/com/flipkart/flux/dao/StatesDAOImpl.java
+++ b/runtime/src/main/java/com/flipkart/flux/dao/StatesDAOImpl.java
@@ -133,8 +133,8 @@ public class StatesDAOImpl extends AbstractDAO<State> implements StatesDAO {
 
     @Override
     @Transactional
-    @SelectDataSource(type = DataSourceType.READ_ONLY, storage = Storage.SHARDED)
-    public List findStatesByDependentEvent(ShardId shardId, String stateMachineId, String eventName) {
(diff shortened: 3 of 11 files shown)


runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,67 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+                                EventData eventData
+    ) throws Exception {
+        try {
+            LoggingUtils.registerStateMachineIdForLogging(machineId);
+            if (eventData.getData() == null || eventData.getName() == null || eventData.getEventSource() == null) {
+                return Response.status(Response.Status.BAD_REQUEST.getStatusCode()).entity(
+                        "Event Data|Name|Source cannot be null.").build();
+            }
+            if (!(eventsDAO.findBySMIdAndName(machineId, eventData.getName()).getStatus() == Event.EventStatus.triggered)) {
+                return Response.status(Response.Status.BAD_REQUEST.getStatusCode()).entity(
```
ykameshrao: IS this BAD_REQUEST or FORBIDDEN/NOT_ACCEPTABLE?

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,67 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+                        "Input event is not in triggered state.").build();
+            }
+            boolean canUpdateEventData = false;
+            /* List holds objects retrieved from States matching input machineId and eventName as [taskId, machineId, status] */
+            List<Object[]> states = parallelScatterGatherQueryHelper.findStatesByDependentEvent(machineId, eventData.getName());
+
+            for (Object[] state : states) {
+                if (state[2] == Status.running) {
+                    return Response.status(Response.Status.BAD_REQUEST.getStatusCode()).entity(
```
ykameshrao: Ditto as above

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,67 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+            if (!(eventsDAO.findBySMIdAndName(machineId, eventData.getName()).getStatus() == Event.EventStatus.triggered)) {
+                return Response.status(Response.Status.BAD_REQUEST.getStatusCode()).entity(
+                        "Input event is not in triggered state.").build();
+            }
+            boolean canUpdateEventData = false;
+            /* List holds objects retrieved from States matching input machineId and eventName as [taskId, machineId, status] */
+            List<Object[]> states = parallelScatterGatherQueryHelper.findStatesByDependentEvent(machineId, eventData.getName());
+
+            for (Object[] state : states) {
+                if (state[2] == Status.running) {
```
ykameshrao: use a switch case

`switch (state) {` 
`    case Status.running: `
`        return NOT_ALLOWED;`
`    default:`
`        canUpdateEventData = true;`
`       break;`
`}`

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,67 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+                return Response.status(Response.Status.BAD_REQUEST.getStatusCode()).entity(
+                        "Event Data|Name|Source cannot be null.").build();
+            }
+            if (!(eventsDAO.findBySMIdAndName(machineId, eventData.getName()).getStatus() == Event.EventStatus.triggered)) {
+                return Response.status(Response.Status.BAD_REQUEST.getStatusCode()).entity(
+                        "Input event is not in triggered state.").build();
+            }
+            boolean canUpdateEventData = false;
+            /* List holds objects retrieved from States matching input machineId and eventName as [taskId, machineId, status] */
+            List<Object[]> states = parallelScatterGatherQueryHelper.findStatesByDependentEvent(machineId, eventData.getName());
```
ykameshrao: Why are Generic not used? Why objects?

client/src/main/java/com/flipkart/flux/client/runtime/FluxRuntimeConnectorHttpImpl.java
```diff
@@ -149,6 +150,24 @@ public void submitScheduledEvent(String name, Object data, String correlationId,
         }
     }
 
+    @Override
+    public void submitEventUpdate(String name, Object data, String correlationId,String eventSource) {
```
ykameshrao: Can data be a generic class?

`public void <T> submitEventUpdate(String name, T data, String corelattionID, String eventSource) {`

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,67 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+                    if (state[2] == Status.errored || state[2] == Status.sidelined) {
+                        this.unsidelineState((String) state[1], (Long) state[0]);
+                    }
+                }
+                return Response.status(Response.Status.ACCEPTED.getStatusCode()).build();
+            }
+        } finally {
+            LoggingUtils.deRegisterStateMachineIdForLogging();
+        }
+        return Response.status(Response.Status.BAD_REQUEST.getStatusCode()).entity("No eligible state dependent on" +
```
ykameshrao: Again, is BAD_REQUEST the right response code?

runtime/src/test/java/com/flipkart/flux/resource/StateMachineResourceTest.java
```diff
@@ -284,6 +284,105 @@ public void testPostScheduledEvent_withTriggerTimeInMilliSeconds() throws Except
...
+            assertThat(state4.getStatus()).isEqualTo(Status.completed);
+            Event event0 = eventsDAO.findBySMIdAndName(smId, "event0");
+            assertThat(event0.getEventData()).isEqualTo("50");
+        } finally {
+            TestWorkflow.shouldFail = false;
+        }
+    }
+
+    @Test
+    public void testEventUpdate_taskRunning() throws Exception {
```
ykameshrao: add tests for other ineligible states too.

client/src/main/java/com/flipkart/flux/client/runtime/FluxRuntimeConnectorHttpImpl.java
```diff
@@ -51,6 +51,7 @@
     public static final int MAX_TOTAL = 200;
     public static final int MAX_PER_ROUTE = 20;
     public static final String EXTERNAL = "external";
+    public static final String EXTERNAL_UPDATE = "externalUpdate";
```
amitkumar94: Why do we need this attribute if we have a specific api for the update purpose..

client/src/main/java/com/flipkart/flux/client/runtime/FluxRuntimeConnectorHttpImpl.java
amitkumar94: We could go generics, but currently everywhere Object notation is used, this refactoring can be added to Tech Debt and picked up in future.

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -393,7 +393,12 @@ public void unsidelineState(String stateMachineId, Long stateId) {
             throw new IllegalStateException("State with the asked id: " + stateId + " not found in stateMachine with id: " + stateMachineId);
         }
 
-        if (askedState.getStatus() == Status.sidelined || askedState.getStatus() == Status.errored) {
+        Set<State> checkExecutableState = new HashSet<>();
```
amitkumar94: Make the parent method non void, send response accordingly if the state is not eligible for un-sidelining.

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -393,7 +393,12 @@ public void unsidelineState(String stateMachineId, Long stateId) {
...
+        checkExecutableState.add(askedState);
+        if(getExecutableStates(checkExecutableState, stateMachineId).isEmpty()) {
+            logger.error("Cannot unsideline state: {}, at least one dependent event is in pending status.", askedState);
```
amitkumar94: ..

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,67 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+    public Response updateEvent(@PathParam("machineId") String machineId,
+                                EventData eventData
+    ) throws Exception {
+        try {
+            LoggingUtils.registerStateMachineIdForLogging(machineId);
+            if (eventData.getData() == null || eventData.getName() == null || eventData.getEventSource() == null) {
+                return Response.status(Response.Status.BAD_REQUEST.getStatusCode()).entity(
+                        "Event Data|Name|Source cannot be null.").build();
+            }
+            if (!(eventsDAO.findBySMIdAndName(machineId, eventData.getName()).getStatus() == Event.EventStatus.triggered)) {
```
amitkumar94: (eventsDAO.findBySMIdAndName(machineId, eventData.getName()).getStatus() != Event.EventStatus.triggered) easier readability ..

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,67 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+                return Response.status(Response.Status.FORBIDDEN.getStatusCode()).entity(
+                        "Input event is not in triggered state.").build();
+            }
+            boolean canUpdateEventData = false;
+            /* List holds objects retrieved from States matching input machineId and eventName as [taskId, machineId, status] */
+            List<Object[]> states = parallelScatterGatherQueryHelper.findStatesByDependentEvent(machineId, eventData.getName());
```
amitkumar94: Why parallel scatter gather, all the information regarding one stateMachine will be one shard..

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,67 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+
+            for (Object[] state : states) {
+                if (state[2] == Status.running) {
+                    return Response.status(Response.Status.NOT_ACCEPTABLE.getStatusCode()).entity(
+                            "EventData update not allowed, one of the states is in running state.").build();
```
amitkumar94: one of the dependent state

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,67 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+                return Response.status(Response.Status.FORBIDDEN.getStatusCode()).entity(
+                        "Input event is not in triggered state.").build();
+            }
+            boolean canUpdateEventData = false;
+            /* List holds objects retrieved from States matching input machineId and eventName as [taskId, machineId, status] */
+            List<Object[]> states = parallelScatterGatherQueryHelper.findStatesByDependentEvent(machineId, eventData.getName());
+
+            for (Object[] state : states) {
+                if (state[2] == Status.running) {
+                    return Response.status(Response.Status.NOT_ACCEPTABLE.getStatusCode()).entity(
```
amitkumar94: dont return from here, if we are using a flag to keep track of execution i.e canUpdateEventData, mark that as false and break

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,67 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+                 * in a customised way for Event. This is temporary for audit purpose, will be adding past Event details
+                 * in a separate effort in future.
+                 * Default values: [{machineId}, 0, 0, null, null, {EventUpdateAudit} String] */
+                String EventUdpateAudit = "Event data updated for event: "+eventData.getName()+" and stateMachineId: "
+                        + machineId;
+                this.auditDAO.create(machineId, new AuditRecord(machineId, Long.valueOf(0), Long.valueOf(0),
+                        null, null, EventUdpateAudit));
+                logger.info("Event data updated for event: {} and stateMachineId: {}", eventData.getName(), machineId);
+                for (Object[] state : states) {
+                    if (state[2] == Status.errored || state[2] == Status.sidelined) {
```
amitkumar94: Shouldn't all the dependent states, be mark as initialized and retriggered, even if they completed succesfully,, as the eventData got updated.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,67 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+                        this.unsidelineState((String) state[1], (Long) state[0]);
```
amitkumar94: Instead of unsidelining , which will again do checks on states, just register in Redriver with 0 scheduled time.

runtime/src/test/java/com/flipkart/flux/resource/StateMachineResourceTest.java
```diff
@@ -284,6 +284,105 @@ public void testPostScheduledEvent_withTriggerTimeInMilliSeconds() throws Except
 
     }
 
+    @Test
+    public void testPostEventUpdate_withNoEventData() throws Exception {
+        String stateMachineDefinitionJson = IOUtils.toString(this.getClass().getClassLoader().getResourceAsStream("state_machine_definition.json"));
```
amitkumar94: There could be cases where event data could be null , in case when they use dummy events to orchestrate their use case. Just check that the field is present, which could be empty., we can discuss this.

client/src/main/java/com/flipkart/flux/client/runtime/FluxRuntimeConnectorHttpImpl.java
akifkhan01: This attribute is used as a default value for "eventSource" column in updated Event DB schema, if user passes null. While updating  eventData via 
    public Event persistEvent(String stateMachineInstanceId, EventData eventData) {
.
event.setEventData(eventData.getData());
event.setEventSource(eventData.getEventSource());
.
}
"eventSource" is required in "eventData" object. And it gives us a separation of all list of events updated externally, filtered on "externalUpdate" eventSource.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
akifkhan01: Done.

client/src/main/java/com/flipkart/flux/client/runtime/FluxRuntimeConnectorHttpImpl.java
amitkumar94: user can just pass "external" as the source, as is the case with external events, we don't need this extra attribute, as the update is via specific api..

akifkhan01: Added it to differentiate eventUpdates. Will remove and use default existing value "external" itself to refrain from adding additional attributes.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
akifkhan01: Done.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
akifkhan01: No more relevant, return statement is removed.

client/src/main/java/com/flipkart/flux/client/runtime/FluxRuntimeConnectorHttpImpl.java
akifkhan01: Can be done as part of refactoring exercise.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
akifkhan01: Can be done as part of refactoring exercise, it's used everywhere as Object now.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
akifkhan01: Not required, removed.

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
akifkhan01: Done. added appropriate response codes.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
akifkhan01: Unsideline also, will eventually register in redriver before execution.

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
amitkumar94: No i meant to raise exception in workflowExecutionController, and send appropriate response from ApiResourceClass.

runtime/src/main/java/com/flipkart/flux/dao/StatesDAOImpl.java
```diff
@@ -131,6 +131,16 @@ public List findStatesByStatus(ShardId shardId, String stateMachineName, Timesta
...
     }
 
-
-
+    @Override
+    @Transactional
+    @SelectDataSource(type = DataSourceType.READ_WRITE, storage = Storage.SHARDED)
+    public List findStatesByDependentEvent(String stateMachineId, String eventName) {
+        Query query;
+        String queryString = "select id, stateMachineId, status from State where stateMachineId = :stateMachineId" +
```
amitkumar94: format this in one single string

runtime/src/main/java/com/flipkart/flux/dao/iface/StatesDAO.java
```diff
@@ -67,4 +67,11 @@
      */
     List findStatesByStatus(ShardId shardId, String stateMachineName, Timestamp fromTime, Timestamp toTime, String taskName, List<Status> statuses);
 
+    /**
+     * Retrieves all states for a particular state-machine-id and like input dependent-event-name.
+     * This query results are specifically used in business logic to make a decision whether to edit EventData or not
```
amitkumar94: Use case of a method need not be mentioned, first line of the paragraph is enough to understand what this method intend to do.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,66 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+            if (eventsDAO.findBySMIdAndName(machineId, eventData.getName()).getStatus() != Event.EventStatus.triggered) {
+                return Response.status(Response.Status.FORBIDDEN.getStatusCode()).entity(
+                        "Input event is not in triggered state.").build();
+            }
+            boolean canUpdateEventData = false;
+            /* List holds objects retrieved from States matching input machineId and eventName as [taskId, machineId, status] */
+            List<Object[]> states = statesDAO.findStatesByDependentEvent(machineId, eventData.getName());
+
+            for (Object[] state : states) {
+                if (state[2] == Status.running) {
```
amitkumar94: switch case as mentioned earlier by Kamesh's review.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,66 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+                    canUpdateEventData = false;
+                    break;
+                } else if (state[2] == Status.initialized || state[2] == Status.errored || state[2] == Status.sidelined) {
+                    canUpdateEventData = true;
+                }
+            }
+            if (canUpdateEventData) {
+                this.workFlowExecutionController.persistEvent(machineId, eventData);
+
+                /* Audit entry in AuditRecord. Though audit record is in alignment with stateId, adding here
```
amitkumar94: Put these steps in a separate function and add these comments over that function.

runtime/src/test/java/com/flipkart/flux/resource/StateMachineResourceTest.java
akifkhan01: Added tests for Cancelled and Completed task status.

runtime/src/test/java/com/flipkart/flux/dao/TestWorkflow.java
```diff
@@ -50,6 +52,14 @@ public Integer dummyTask(String integer) {
                 e.printStackTrace();
             }
         }
+        if(shouldCancel) {
+            try {
+                throw new FluxCancelPathException();
+            } catch (Exception e){
+                System.out.println("Failed to cancel path.");
```
amitkumar94: This will be always be caught, check if the exception raised is not an instance of FluxCancelPath, something like this.
```
if( !e instanceof FluxCancelPathException){
 print ...
}
```

runtime/src/test/resources/updated_event_data.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "event0",
+  "type": "java.lang.String",
+  "data": "50",
+  "eventSource": "externalUpdate"
```
amitkumar94: eventSource: external

runtime/src/test/java/com/flipkart/flux/resource/StateMachineResourceTest.java
```diff
@@ -284,6 +284,145 @@ public void testPostScheduledEvent_withTriggerTimeInMilliSeconds() throws Except
...
+        final HttpResponse<String> smCreationResponse = Unirest.post(STATE_MACHINE_RESOURCE_URL).header(
+                "Content-Type", "application/json").body(stateMachineDefinitionJson).asString();
+        Event event = eventsDAO.findBySMIdAndName(smCreationResponse.getBody(), "event0");
+        assertThat(event.getStatus()).isEqualTo(Event.EventStatus.pending);
+
+        /* Make the task fail, eventually sidelined. */
+        TestWorkflow.shouldFail = true;
+        try {
+            /* Since event0 is in pending state, updateEvent should fail. */
+            String eventJson0 = IOUtils.toString(this.getClass().getClassLoader().getResourceAsStream(
```
amitkumar94: correct eventData should fail as well as task in pending state. Test with valid eventData.

runtime/src/main/java/com/flipkart/flux/dao/StatesDAOImpl.java
akifkhan01: It's formatted, please view it as a whole file.

runtime/src/main/java/com/flipkart/flux/dao/iface/StatesDAO.java
akifkhan01: Done.

runtime/src/test/resources/updated_event_data.json
akifkhan01: Done.

runtime/src/test/java/com/flipkart/flux/dao/TestWorkflow.java
akifkhan01: Done.

runtime/src/test/java/com/flipkart/flux/resource/StateMachineResourceTest.java
```diff
@@ -284,6 +284,145 @@ public void testPostScheduledEvent_withTriggerTimeInMilliSeconds() throws Except
...
+    private void testEventUpdate_IneligibleTaskStatus_Util(HttpResponse<String> smCreationResponse) throws Exception {
+        String eventJson1 = IOUtils.toString(this.getClass().getClassLoader().getResourceAsStream("event_data.json"));
+        String eventJson2 = IOUtils.toString(this.getClass().getClassLoader().getResourceAsStream(
+                "updated_event_data.json"));
+
+        final HttpResponse<String> eventPostResponse1 = Unirest.post(
+                STATE_MACHINE_RESOURCE_URL + SLASH + smCreationResponse.getBody() + "/context/events").header(
+                "Content-Type", "application/json").body(eventJson1).asString();
+            /* Try update Event while task is in running state */
+        final HttpResponse<String> eventPostResponse2 = Unirest.post(
```
amitkumar94: Is this comment still valid, when this helper method called by multiple tests.

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
akifkhan01: Done.

runtime/src/test/java/com/flipkart/flux/resource/StateMachineResourceTest.java
akifkhan01: Removed.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -299,6 +299,66 @@ public Response submitEvent(@PathParam("machineId") String machineId,
...
+                for (Object[] state : states) {
+                    if (state[2] == Status.initialized || state[2] == Status.errored || state[2] == Status.sidelined) {
+                        return this.workFlowExecutionController.unsidelineState((String) state[1], (Long) state[0]);
+                    }
+                }
+            }
+        } finally {
+            LoggingUtils.deRegisterStateMachineIdForLogging();
+        }
+        return Response.status(Response.Status.NOT_FOUND.getStatusCode()).entity("No eligible state dependent on" +
```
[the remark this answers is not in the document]
akifkhan01: Done.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
akifkhan01: Done.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
akifkhan01: As per design, only triggered events with at least one of the next/immediate dependent state(s) in errored/sidelined/initialized state and none of these states in running state can be updated. 
If we use switch case, status can be categorised in 3:
1. Running - Update not allowed irrespective of what is the status of other states.
2. [ Completed, Cancelled, Unsidelined ] - Update not allowed for this state but event data can be updated if at least one of the other dependent states is [ sidelined, errored, initialized ] and none of them is in [running] status.
3. [ sidelined, errored, initialized ] - Event Data update allowed if none of the dependent states is in [running] status.

Since, states can be iterated in the loop/switch case in any order, maintaining results with only one flag 'canUpdateEventData' doesn't help. Also adding flag for encountered 'running' state doesn't help.

Consider one failure scenario, Say dependent states are as: 
[ T1, initialized], [T2, completed] - Event data can be updated.
    Loop iteration order:
    1. [T1, T2] - Will not update (cannot be handled in switch unless we make it more complex by adding few more flags).
    2. [T2, T1] - Will update.

For switch case defined as  :
for (Object[] state : states) {
                Status status = (Status)state[2];
                switch (status) {
                    case running:
                        encounteredRunningState = true;
                        canUpdateEventData = false;
                        break;
                    case initialized:
                    case errored:
                    case sidelined:
                        canUpdateEventData = true;
                        break;
                    case completed:
                    case cancelled:
                        canUpdateEventData = false;
                        break;
                }
                if(encounteredRunningState) {
                    break;
                }
            }

However, in existing if-else, it's handled if 'running' status encountered, mark flag to false and break (update invalid) else 'initilized|sidelined|errored', mark flag to true but don't break for loop because we may encounter 'running' state later.

PR #79, opened 2018-08-16
updated redriver deletion to bulk delete, improved redriver polling a…

amitkumar94:
…nd async redrive through pool, and all tests passing

### DIFF
--- a/scheduler/src/main/java/com/flipkart/flux/redriver/service/MessageManagerService.java
+++ b/scheduler/src/main/java/com/flipkart/flux/redriver/service/MessageManagerService.java
@@ -89,14 +89,15 @@ public class MessageManagerService implements Initializable {
     public void initialize() {
         scheduledDeletionService.scheduleAtFixedRate(() -> {
             try {
-                logger.info("Running Deletion job, deleting {} messages", Math.min(batchSize, messagesToDelete.size()));
+                logger.info("Running Deletion job, trying deleting {} messages", Math.min(batchSize, messagesToDelete.size()));
                 SmIdAndTaskIdPair currentMessageIdToDelete = null;
                 List messageIdsToDelete = new ArrayList<SmIdAndTaskIdPair>(batchSize);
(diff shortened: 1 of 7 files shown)


scheduler/src/main/java/com/flipkart/flux/redriver/service/MessageManagerService.java
```diff
@@ -89,14 +89,15 @@ public void scheduleForRemoval(String stateMachine, Long taskId) {
     public void initialize() {
         scheduledDeletionService.scheduleAtFixedRate(() -> {
             try {
-                logger.info("Running Deletion job, deleting {} messages", Math.min(batchSize, messagesToDelete.size()));
+                logger.info("Running Deletion job, trying deleting {} messages", Math.min(batchSize, messagesToDelete.size()));
```
akifkhan01: Can be written as "Running Deletion job, trying to delete {} messages"...

scheduler/src/test/java/com/flipkart/flux/redriver/scheduler/RedriverServiceTest.java
```diff
@@ -91,19 +88,6 @@ public void testLargeNoOfRedriveMessages_allShouldBeRedrived() throws Exception
...
-        when(messageManagerService.retrieveOldest(0, batchSize)).
-                thenReturn(Arrays.asList(new ScheduledMessage(1l, "sample-state-machine-uuid", now + 9000),
-                        new ScheduledMessage(2l, "sample-state-machine-uuid", now + 10000)));
-
-        redriverService.start();
-
-        Thread.sleep(1000);
-        verifyZeroInteractions(redriverRegistry);
-    }
-
```
akifkhan01: Why is this test case removed ?

amitkumar94: will add a integration test to cover this unit one.

PR #80, opened 2018-09-28
Duplicate event drop

amitkumar94:
Most are reformatting changes.


runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -323,8 +314,16 @@ public Event persistEvent(String stateMachineInstanceId, EventData eventData) {
...
     @SelectDataSource(type = DataSourceType.READ_WRITE, storage = Storage.SHARDED)
-    public void updateTaskStatus(String machineId, Long stateId, ExecutionUpdateData executionUpdateData) {
+    public void updateTaskStatus(String machineId, Long stateId, ExecutionUpdateData executionUpdateData) throws
+            IllegalStateException {
         com.flipkart.flux.domain.Status updateStatus = null;
+        State state = statesDAO.findById(machineId, stateId);
+        if (state.getStatus().equals(Status.completed) || state.getStatus().equals(Status.cancelled)) {
+            logger.warn("State {} from smid: {} is already completed/cancelled. Rejected the updateState request",
+                    stateId, machineId);
+            throw new IllegalStateException("Task is already completed, rejecting any further updates on it.");
```
ykameshrao: let the error exception also say 'Task is already completed/cancelled'

runtime/src/main/java/com/flipkart/flux/controller/WorkFlowExecutionController.java
```diff
@@ -128,43 +133,23 @@ public WorkFlowExecutionController(EventsDAO eventsDAO, StateMachinesDAO stateMa
         return initialStates;
     }
 
-    /**
-     * Updates task status and retrieves the states which are dependant on this event and starts the execution of eligible states (whose all dependencies are met).
-     * @param stateMachine
-     * @param eventAndExecutionData
-     */
-    public void updateTaskStatusAndPostEvent(StateMachine stateMachine, EventAndExecutionData eventAndExecutionData) {
```
ykameshrao: Why is this code getting deleted?

amitkumar94: unused code.

PR #81, opened 2018-11-22
exposed task execution metrics

amitkumar94:


task/src/main/java/com/flipkart/flux/impl/task/AkkaTask.java
```diff
@@ -155,10 +157,12 @@ public void onReceive(Object message) throws Exception {
                         logger.info("State machine: {} task: {} execution time: {}ms event/status submission time: {}ms", taskAndEvent.getStateMachineId(), taskAndEvent.getTaskId(), (endTime - startTime), (System.currentTimeMillis() - endTime));
 
                     } catch (HystrixRuntimeException hre) {
+                        context.close();
                         FailureType ft = hre.getFailureType();
                         // we signal a timeout for any of Timeout, ThreadPool Rejection or Short-Circuit - all of these may go through with time and retry
                         if (ft.equals(FailureType.REJECTED_THREAD_EXECUTION) || ft.equals(FailureType.SHORTCIRCUIT) || ft.equals(FailureType.TIMEOUT)) {
                             // update flux runtime with task outcome as timeout
+                            if(ft.equals(FailureType.TIMEOUT))
```
akifkhan01: Why is this required ? If we are doing this then above condition 
"if (ft.equals(FailureType.REJECTED_THREAD_EXECUTION) || ft.equals(FailureType.SHORTCIRCUIT) || ft.equals(FailureType.TIMEOUT))" comes down to "if(ft.equals(FailureType.TIMEOUT))" itself.

akifkhan01:
Reviewed. I have added a query, otherwise it looks good.

PR #84, opened 2019-02-04
Modified response code for ConstraintViolation

akifkhan01:
In case of constraint violation, changed response code to 500
(INTERNAL_SERVER_ERROR) if it is not for duplicate State Machine Id.
For duplciate state machine id, it sends 409 (CONFLICT) as response.

All tests passing.

### DIFF
--- a/runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
+++ b/runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
@@ -138,7 +138,7 @@ public class StateMachineResource {
         if (stateMachineDefinition == null)
             throw new IllegalRepresentationException("State machine definition is empty");
 
-        StateMachine stateMachine = null, stateMachine1 = null;
+        StateMachine stateMachine = null;
 
 
         final String stateMachineInstanceId;
--- a/runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
+++ b/runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
@@ -160,14 +160,13 @@ public class StateMachineResource {
                     append(".started").toString());
         } catch (ConstraintViolationException ex) {
             //In case of Duplicate correlation key, return http code 409 conflict
-            stateMachine1 = stateMachinesDAO.findById(stateMachineInstanceId);
-            if (stateMachine1 != null) {
+            if (ex.getCause().getMessage().contains("Duplicate entry")) {
                 return Response.status(Response.Status.CONFLICT.getStatusCode()).entity(
                         ex.getCause() != null ? ex.getCause().getMessage() : null).build();
 
             }
             logger.error("Constraint Violation during creating or initiating StateMachine" +
-                    " with id {} {}", stateMachineInstanceId, ex.getStackTrace());
+                    " with id {} {} {}", stateMachineInstanceId, ex.getCause().getMessage(), ex.getStackTrace());
             return Response.status(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode()).entity(
                     ex.getCause() != null ? ex.getCause().getMessage() : null).build();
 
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -160,11 +159,22 @@ public Response createStateMachine(StateMachineDefinition stateMachineDefinition
                     append(stateMachine.getName()).
                     append(".started").toString());
         } catch (ConstraintViolationException ex) {
-            //in case of Duplicate correlation key, return http code 409 conflict
-            return Response.status(Response.Status.CONFLICT.getStatusCode()).entity(ex.getCause() != null ? ex.getCause().getMessage() : null).build();
+            //In case of Duplicate correlation key, return http code 409 conflict
+            stateMachine1 = stateMachinesDAO.findById(stateMachineInstanceId);
```
amitkumar94: Check when ex.getMessage() contains DuplicatePrimary Key else throw 5xx, don't make a find call

akifkhan01: Done.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -139,7 +138,7 @@ public Response createStateMachine(StateMachineDefinition stateMachineDefinition
         if (stateMachineDefinition == null)
             throw new IllegalRepresentationException("State machine definition is empty");
 
-        StateMachine stateMachine = null;
+        StateMachine stateMachine = null, stateMachine1 = null;
```
amitkumar94: stateMachine1 is no longer needed

akifkhan01: Removed.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -160,11 +159,23 @@ public Response createStateMachine(StateMachineDefinition stateMachineDefinition
                     append(stateMachine.getName()).
                     append(".started").toString());
         } catch (ConstraintViolationException ex) {
-            //in case of Duplicate correlation key, return http code 409 conflict
-            return Response.status(Response.Status.CONFLICT.getStatusCode()).entity(ex.getCause() != null ? ex.getCause().getMessage() : null).build();
+            //In case of Duplicate correlation key, return http code 409 conflict
+            if(ex.getCause() != null) {
+                if (ex.getCause().getMessage().contains("Duplicate entry")) {
```
ykameshrao: ex.getClause().getMessage().toLowerCase().contains("duplicate entry") 

make it case agnostic.

ykameshrao: can you instead do this in the underlying DAO layer and throw relevant exception from their? DATA layer details are leaking to Resource layer which is exactly what DAO was supposed to abstract out.

akifkhan01: > ex.getClause().getMessage().toLowerCase().contains("duplicate entry")
> 
> make it case agnostic.

Done, thanks for pointing out.

runtime/src/main/java/com/flipkart/flux/resource/StateMachineResource.java
```diff
@@ -160,11 +159,23 @@ public Response createStateMachine(StateMachineDefinition stateMachineDefinition
                     append(stateMachine.getName()).
                     append(".started").toString());
         } catch (ConstraintViolationException ex) {
-            //in case of Duplicate correlation key, return http code 409 conflict
-            return Response.status(Response.Status.CONFLICT.getStatusCode()).entity(ex.getCause() != null ? ex.getCause().getMessage() : null).build();
+            //In case of Duplicate correlation key, return http code 409 conflict
+            if(ex.getCause() != null) {
+                if (ex.getCause().getMessage().toLowerCase().contains("duplicate entry")) {
+                    return Response.status(Response.Status.CONFLICT.getStatusCode()).entity(
+                            ex.getCause() != null ? ex.getCause().getMessage() : null).build();
```
ykameshrao: isn't this null check redundant? you already checked at line 163.

akifkhan01: Removed redundant check.

PR #106, opened 2022-08-12
Changes to enable/disable auditing to database via config

atulsm:
Details: https://confluence.fkinternal.com/pages/viewpage.action?spaceKey=SF&title=Flux+Query+Analysis

This PR will enable the capability to control audit behavior with a flag enableAuditRecord. Primarily intended to improve the overall throughput and better data management while we lose the capability to see audit info in fsmview.

Results from loadtest => About 14% overall throughput improvement in local env.

1000 WF's, Audit Enabled (Default):
Time: All completed in 28 seconds

1000 WF's, Audit Disabled:
Time: All completed in 24 seconds


bageshwar:
@atulsm Can you share an externally accessible link ?

